PR #1477, opened 2025-03-31
Qualcomm AI Engine Direct - Enable BroadcastTo OP

chuntl:
Summary:
- Enable op builder for broadcast to op
- Add 2 mlir model for different data type

### DIFF
--- a/litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
+++ b/litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
@@ -1,0 +1,144 @@
+// Copyright (c) Qualcomm Innovation Center, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.h"
+
+#include <cstdint>
+#include <numeric>
+#include <vector>
+
+#include "litert/vendors/qualcomm/core/builders/op_builder.h"
+#include "litert/vendors/qualcomm/core/tensor_pool.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "third_party/qairt/latest/include/QNN/QnnOpDef.h"
+
+namespace qnn {
+namespace {
+std::vector<std::uint32_t> GetStaticTensorDimention(
+    const std::vector<std::uint32_t>& input_dimensions,
+    const std::vector<std::uint32_t>& output_dimensions) {
+  std::uint32_t input_size = input_dimensions.size();
+  std::uint32_t output_size = output_dimensions.size();
+  std::vector<std::uint32_t> final_dimensions;
+
+  if (input_size < output_size) {
+    int padding_size = output_size - input_size;
+    final_dimensions = std::vector<std::uint32_t>(padding_size, 1);
+    final_dimensions.insert(final_dimensions.end(), input_dimensions.begin(),
+                            input_dimensions.end());
+  } else {
+    final_dimensions = input_dimensions;
+  }
+
+  for (std::size_t i = 0; i < output_size; ++i) {
+    final_dimensions[i] =
+        (output_dimensions[i] > final_dimensions[i]) ? output_dimensions[i] : 1;
+  }
+
+  return final_dimensions;
+}
+
+template <typename T>
+TensorWrapper& CreateStaticTensor(TensorPool& tensor_pool,
+                                  const Qnn_DataType_t data_type,
+                                  TensorWrapper& input, TensorWrapper& output,
+                                  bool is_quant) {
+  std::vector<std::uint32_t> static_dims =
+      GetStaticTensorDimention(input.GetDims(), output.GetDims());
+  std::uint32_t static_size =
+      std::accumulate(static_dims.begin(), static_dims.end(), 1,
+                      std::multiplies<std::uint32_t>());
+
+  T static_value{0};
+  QuantizeParamsWrapperVariant quant_param{};
+  if (is_quant) {
+    auto input_quant_param =
+        std::get<ScaleOffsetQuantizeParamsWrapper>(input.GetQuantParams());
+    quant_param.emplace<ScaleOffsetQuantizeParamsWrapper>(input_quant_param);
+    static_value = -input_quant_param.GetZeroPoint();
+  }
+  std::vector<T> static_data(static_size, static_value);
+
+  return tensor_pool.CreateStaticTensor(data_type, quant_param, static_dims,
+                                        sizeof(T) * static_size,
+                                        static_data.data());
+}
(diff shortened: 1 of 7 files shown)


litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+  const char* qnn_op = nullptr;
+  if (inputs[0].get().GetDataType() == QNN_DATATYPE_BOOL_8) {
+    qnn_op = QNN_OP_ELEMENT_WISE_OR;
+  } else {
+    qnn_op = QNN_OP_ELEMENT_WISE_ADD;
+  }
+
+  auto& broadcast_op = CreateOpWrapper(res, qnn_op);
+  broadcast_in = &(inputs[0].get());
+  broadcast_op.AddInputTensor(*broadcast_in);
```
weilhuan-quic: We don't need `broadcast_in` because it is only used in one place. 

```suggestion
  broadcast_op.AddInputTensor(inputs[0]);
```

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+    }
+    default: {
+      QNN_LOG_ERROR("Unsupported QNN data type when creating static tensor");
+      break;
+    }
+  }
+
+  broadcast_op.AddOutputTensor(outputs[0]);
+
+  // TODO: fused activation
```
weilhuan-quic: There is no fused activation for BroadcastTo.

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+    final_dims[i] = (output[i] > input[i]) ? output[i] : 1;
+  }
+  return final_dims;
+}
+
+template <typename T>
+TensorWrapper& CreateStaticTensor(TensorPool& tensor_pool,
+                                  const Qnn_DataType_t data_type,
+                                  qnn::TensorWrapperRef output,
+                                  qnn::TensorWrapperRef input) {
```
weilhuan-quic: You can directly pass reference instead of reference wrapper.

```suggestion
                                  TensorWrapper& output,
                                  TensorWrapper& input) {
```

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "litert/vendors/qualcomm/core/builders/op_builder.h"
+#include "litert/vendors/qualcomm/core/tensor_pool.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "third_party/qairt/latest/include/QNN/QnnOpDef.h"
+
+namespace qnn {
+std::vector<std::uint32_t> GetStaticTensorDimention(
+    const std::vector<std::uint32_t>& output,
+    const std::vector<std::uint32_t>& input) {
```
weilhuan-quic: Better naming.

```suggestion
    const std::vector<std::uint32_t>& output_dimensions,
    const std::vector<std::uint32_t>& input_dimensions) {
```

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+  std::vector<std::uint32_t> final_dims(output.size());
```
weilhuan-quic: Current assumption is that the dimension of input and output tensors are the same. But according the definition, we need to handle the tensor with fewer axes. 

> When broadcasting, if a tensor has fewer axes than necessary its shape is padded on the left with ones. So this gives the same result as the previous example:


https://www.tensorflow.org/api_docs/python/tf/broadcast_to

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+  if (data_type == QNN_DATATYPE_UFIXED_POINT_8 ||
+      data_type == QNN_DATATYPE_SFIXED_POINT_8 ||
+      data_type == QNN_DATATYPE_UFIXED_POINT_16 ||
+      data_type == QNN_DATATYPE_SFIXED_POINT_16) {
+    quant_param.emplace<ScaleOffsetQuantizeParamsWrapper>(1, 0);
+  }
+
+  return tensor_pool.CreateStaticTensor(data_type, quant_param, static_dims,
+                                        sizeof(T) * static_size,
+                                        static_data.data());
```
weilhuan-quic, line 65: Need to resize `static_data` into the multiply of `static_dims`, otherwise we will access the invalid memory.

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+      GetStaticTensorDimention(output.get().GetDims(), input.get().GetDims());
+  std::uint32_t static_size =
+      std::accumulate(static_dims.begin(), static_dims.end(), 1,
+                      std::multiplies<std::uint32_t>());
+
+  QuantizeParamsWrapperVariant quant_param = QNN_QUANTIZE_PARAMS_INIT;
+  if (data_type == QNN_DATATYPE_UFIXED_POINT_8 ||
+      data_type == QNN_DATATYPE_SFIXED_POINT_8 ||
+      data_type == QNN_DATATYPE_UFIXED_POINT_16 ||
+      data_type == QNN_DATATYPE_SFIXED_POINT_16) {
```
weilhuan-quic: `data_type` is already checked outside, can use a flag like `is_quant` as the function parameter instead of checking again.

litert/vendors/qualcomm/core/builders/broadcast_to_op_builder.cc
```diff
@@ -0,0 +1,119 @@
...
+#include <vector>
+
+#include "litert/vendors/qualcomm/core/builders/op_builder.h"
+#include "litert/vendors/qualcomm/core/tensor_pool.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "third_party/qairt/latest/include/QNN/QnnOpDef.h"
+
+namespace qnn {
+std::vector<std::uint32_t> GetStaticTensorDimention(
```
weilhuan-quic, line 18: Can put `GetStaticTensorDimention` and `CreateStaticTensor` in the anonymous namespace.

weilhuan-quic (approved):

YunanAZ (approved):

YunanAZ (approved):

PR #1582, opened 2025-04-05
Qualcomm AI Engine Direct - Fused Activation

weilhuan-quic:
1. Support Relu, ReluN1To1, Relu6, Tanh
2. If activation is not none, replace output tensor before op builder and add fused activation node after op builder.
3. Support fused activation for the covered ops, remove TDOO in op builder.

### DIFF
--- a/litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
+++ b/litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
@@ -267,11 +267,11 @@ LiteRtStatus ConvertTensor(const litert::Tensor& litert_tensor,
   return kLiteRtStatusOk;
 }
 
-LiteRtStatus ConvertOp(
-    const litert::Op& litert_op, ::qnn::TensorPool& tensor_pool,
-    const std::vector<::qnn::TensorWrapperRef>& input_tensors,
-    const std::vector<::qnn::TensorWrapperRef>& output_tensors,
-    std::vector<::qnn::OpWrapper>& op_wrappers) {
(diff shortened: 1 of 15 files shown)


litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
```diff
@@ -557,9 +590,14 @@ LiteRtStatus ConvertOp(
 
       ::qnn::PaddingType qnn_padding;
       LITERT_RETURN_IF_ERROR(ConvertPaddingType(padding, qnn_padding));
+
+      auto& activation_output = ::qnn::ReplaceOutputTensorForFusedActivation(
```
ghost, line 594: NIT:
`ReplaceOutputTensorForFusedActivation` is a bit confused since `BuildXXXXOp` does not use the output. 
Developer might be confused about the order when using these functions.
Probably we need to change the function name or refactor a bit.

weilhuan-quic: Agree, I think this PR is not required to merge soon and let me think about better way to handle this.

weilhuan-quic: Add comments for `ReplaceOutputTensorForFusedActivation`. If developers use this function in the wrong order, they will encounter finalization failure. Therefore, I think we can keep the current implementation and change it if more developers complain.

chunhsue (approved):

YunanAZ (approved):

PR #1751, opened 2025-04-16
Qualcomm AI Engine Direct - Reduce uint16/int16 convert time

jiunkaiy:


litert/vendors/qualcomm/core/wrappers/tensor_wrapper.cc
```diff
@@ -69,10 +69,14 @@ TensorWrapper::TensorWrapper() = default;
...
-      dimentions_{dimentions},
-      quantize_params_{quantize_params} {
+    const std::vector<std::uint32_t>& dimentions,
+    absl::string_view tensor_name)
+    : dimentions_{dimentions}, quantize_params_{quantize_params} {
+  if (tensor_name.empty()) {
+    name_ = std::to_string(id);
+  } else {
+    name_ = tensor_name;
+  }
```
weilhuan-quic: We can use both `id` and `tensor_name` like:
```
name_ = tensor_name;
name_ += "_id_" + std::to_string(id);
```

Then the user of `TensorWrapper` can see the effect of the argument they provide.

litert/vendors/qualcomm/core/tensor_pool.h
```diff
@@ -24,12 +25,14 @@ class TensorPool {
   TensorWrapper& CreateInputTensor(
       Qnn_DataType_t data_type,
       const QuantizeParamsWrapperVariant& quant_params,
-      const std::vector<std::uint32_t>& dimentions);
+      const std::vector<std::uint32_t>& dimentions,
+      absl::string_view tensor_name);
```
weilhuan-quic, line 29: Nit: add comments to indicate why we only need to provide `tensor_name` for input and output tensor. And let me open a internal ticket to support readable tensor name for all tensor types.

litert/vendors/qualcomm/core/utils/miscs.cc
```diff
@@ -16,13 +34,25 @@ void ConvertDataFromInt16toUInt16(absl::Span<const std::int16_t> src,
...
+#else
+  for (auto& data : src) {
+    data += kUint16ZeroPoint;
+  }
+#endif
+}
+
+void ConvertDataFromUInt16toInt16(absl::Span<std::uint16_t> src) {
+#ifdef __ANDROID__
+  neon_xor(reinterpret_cast<std::uint16_t*>(src.data()), src.size());
```
weilhuan-quic: I think we don't need `reinterpret_cast` here.

litert/vendors/qualcomm/dispatch/litert_dispatch_invocation_context.cc
```diff
@@ -263,10 +263,14 @@ Expected<void> LiteRtDispatchInvocationContextT::AttachBuffer(
 
 Expected<void> LiteRtDispatchInvocationContextT::Execute() {
   for (int i = 0; i < inputs_.size(); ++i) {
+    if (strncmp(inputs_[i].GetName(), "kv_cache", 8) == 0) {
```
weilhuan-quic: If `GetName` return a `absl::string_view`, we can use its member function `starts_with`. I think it is more readable and reilable..

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.cc
jiunkaiy: Changed to `name_ = std::string(tensor_name) + "_id_" + std::to_string(id);` Thanks!

litert/vendors/qualcomm/core/tensor_pool.h
jiunkaiy: I have added some comments. Thanks.

litert/vendors/qualcomm/core/utils/miscs.cc
jiunkaiy: I have changed all data conversion between int16 anb uint16 to the new function `ToggleMsb`.

litert/vendors/qualcomm/dispatch/litert_dispatch_invocation_context.cc
jiunkaiy: I can't apply `starts_with` even with `absl`. This is a feature introduced in C++20.
For now, I will continue using `strncmp` to maintain low comparison overhead.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -31,11 +34,30 @@ float Dequantize(const T val, const float scale, const int32_t zero_point) {
...
-                                  std::vector<std::uint16_t>& dst);
-
-void ConvertDataFromUInt16toInt16(absl::Span<const std::uint16_t> src,
-                                  std::vector<std::int16_t>& dst);
+template <typename T>
+using EnableIfInt16OrUint16Ptr = std::enable_if_t<
+    std::is_same_v<T, int16_t*> || std::is_same_v<T, uint16_t*>, bool>;
+// Converts data between quantized UINT16 and INT16 formats.
+template <typename T, EnableIfInt16OrUint16Ptr<T> = true>
+void ToggleMsb(T src, size_t tensor_size) {
```
weilhuan-quic: I guess `Msb` means most significant bit, but it may not obvious for others.

Also change the name of the second parameter. 

```suggestion
void ToggleMostSignificantBit(T src, size_t size) {
```

weilhuan-quic (approved):

weilhuan-quic (approved):

PR #1797, opened 2025-04-21
Qualcomm AI Engine Direct - Fix MaxPool2D in ResNet50

chuntl:
Summary:
- Fix the calculation function for padding used in MaxPool2D

The attributes of MaxPool2D here are:
| Property                    | Value |
|-----------------------------|-------|
| **filter_height**           | 3 : i32 |
| **filter_width**            | 3 : i32 |
| **padding**                 | VALID |
| **stride_h**                | 2 : i32 |
| **stride_w**                | 2 : i32 |
| **fused_activation_function** | NONE |


chuntl:
After fixing with this PR, the result is same as tflite side. The value of padding here is:

Before fix:
image

After fix (Correct):
image

YunanAZ (approved):

PR #2446, opened 2025-06-06
Qualcomm AI Engine Direct - Dump QNN IR JSON graph.

jiunkaiy:
Summary:
- Implement the option to dump Qnn JSON.
- Ensure compatibility with the Qnn converter and ORT Qnn JSON format.
- Introduce unit tests for DumpQnnJson and the corresponding option.

```
bazel-bin/litert/tools/apply_plugin_main -cmd apply --libs bazel-bin/litert/vendors/qualcomm/compiler --soc_model SM8650 --soc_manufacturer Qualcomm  --model /local/mnt/workspace/jiunkaiy/LiteRT/models/model_gemma3/static_a16w4-full-int_qpa_quantized_gemma3_npu_f32_ekv1280.tflite --qualcomm_ir_json_dir ./tmp123/
```
image

image

### DIFF
--- a/litert/c/options/litert_qualcomm_options.cc
+++ b/litert/c/options/litert_qualcomm_options.cc
@@ -20,6 +20,7 @@
 #include <cstdint>
 #include <memory>
 #include <vector>
+#include <string>
 
 #include "absl/strings/string_view.h"  // from @com_google_absl
 #include "litert/c/litert_common.h"
@@ -35,6 +36,7 @@ struct LiteRtQualcommOptionsT {
   LiteRtQualcommOptionsHtpPerformanceMode htp_performance_mode =
       kLiteRtQualcommHtpPerformanceModeDefault;
   std::vector<std::int32_t> dump_tensor_ids;
+  std::string ir_json_dir;
 };
 
 LiteRtStatus LiteRtQualcommOptionsCreate(LiteRtOpaqueOptions* options) {
@@ -241,6 +243,28 @@ LiteRtStatus LiteRtQualcommOptionsGetProfiling(
   }
 
   *profiling = options->profiling;
+
+  return kLiteRtStatusOk;
+}
+
+LiteRtStatus LiteRtQualcommOptionsSetIrJsonDir(LiteRtQualcommOptions options,
+                                               const char* ir_json_dir) {
+  if (options == nullptr) {
+    return kLiteRtStatusErrorInvalidArgument;
+  }
+
+  options->ir_json_dir = ir_json_dir;
+
+  return kLiteRtStatusOk;
+}
+
+LiteRtStatus LiteRtQualcommOptionsGetIrJsonDir(LiteRtQualcommOptions options,
+                                               const char** ir_json_dir) {
+  if (options == nullptr || ir_json_dir == nullptr) {
+    return kLiteRtStatusErrorInvalidArgument;
+  }
+
+  *ir_json_dir = options->ir_json_dir.data();
 
   return kLiteRtStatusOk;
 }
--- a/litert/c/options/litert_qualcomm_options_test.cc
+++ b/litert/c/options/litert_qualcomm_options_test.cc
@@ -153,6 +153,23 @@ TEST(LiteRtQualcommOptionsTest, Profiling) {
   LiteRtDestroyOpaqueOptions(options);
 }
 
+TEST(LiteRtQualcommOptionsTest, IrJsonDir) {
+  LiteRtOpaqueOptions options;
+  LITERT_ASSERT_OK(LiteRtQualcommOptionsCreate(&options));
+
+  LiteRtQualcommOptions qualcomm_options;
+  LITERT_ASSERT_OK(LiteRtQualcommOptionsGet(options, &qualcomm_options));
+
+  LITERT_ASSERT_OK(LiteRtQualcommOptionsSetIrJsonDir(qualcomm_options, "tmp/"));
+
+  const char* ir_json_dir;
+  LITERT_ASSERT_OK(
+      LiteRtQualcommOptionsGetIrJsonDir(qualcomm_options, &ir_json_dir));
+  EXPECT_STREQ(ir_json_dir, "tmp/");
+
+  LiteRtDestroyOpaqueOptions(options);
+}
+
 TEST(LiteRtQualcommOptionsTest, DumpTensorIds) {
   LiteRtOpaqueOptions options;
   LITERT_ASSERT_OK(LiteRtQualcommOptionsCreate(&options));
@@ -213,6 +230,10 @@ TEST(QualcommOptionsTest, CppApi) {
   for (size_t i = 0; i < kDumpTensorIds.size(); i++) {
     EXPECT_EQ(ids[i], kDumpTensorIds[i]);
   }
+
+  EXPECT_EQ(options->GetIrJsonDir(), "");
+  options->SetIrJsonDir("tmp");
+  EXPECT_EQ(options->GetIrJsonDir(), "tmp");
 }
 
 TEST(QualcommOptionsTest, FindFromChain) {
--- a/litert/cc/options/litert_qualcomm_options.cc
+++ b/litert/cc/options/litert_qualcomm_options.cc
@@ -136,6 +136,17 @@ std::vector<std::int32_t> QualcommOptions::GetDumpTensorIds() {
   return dump_ids;
 }
 
+void QualcommOptions::SetIrJsonDir(const std::string& ir_json_dir) {
+  internal::AssertOk(LiteRtQualcommOptionsSetIrJsonDir, Data(),
+                     ir_json_dir.c_str());
+}
+
+absl::string_view QualcommOptions::GetIrJsonDir() {
+  const char* ir_json_dir;
+  internal::AssertOk(LiteRtQualcommOptionsGetIrJsonDir, Data(), &ir_json_dir);
+  return absl::string_view(ir_json_dir);
+}
+
 Expected<QualcommOptions> QualcommOptions::Create(OpaqueOptions& options) {
   const auto id = options.GetIdentifier();
   if (!id || *id != Discriminator()) {
--- a/litert/cc/options/litert_qualcomm_options.h
+++ b/litert/cc/options/litert_qualcomm_options.h
@@ -23,6 +23,7 @@
 #include "litert/c/options/litert_qualcomm_options.h"
 #include "litert/cc/litert_expected.h"
 #include "litert/cc/litert_opaque_options.h"
+#include "absl/strings/string_view.h"  // from @com_google_absl
 
 namespace litert::qualcomm {
 
@@ -60,6 +61,9 @@ class QualcommOptions : public OpaqueOptions {
 
   void SetDumpTensorIds(const std::vector<std::int32_t>& ids);
   std::vector<std::int32_t> GetDumpTensorIds();
+
+  void SetIrJsonDir(const std::string& profiling);
+  absl::string_view GetIrJsonDir();
 
  private:
   LiteRtQualcommOptions Data() const;
--- a/litert/tools/flags/vendors/qualcomm_flags.cc
+++ b/litert/tools/flags/vendors/qualcomm_flags.cc
@@ -92,8 +92,8 @@ ABSL_FLAG(bool, qualcomm_use_qint16_as_quint16, false,
           "quantized uin16 model.");
 
 ABSL_FLAG(LiteRtQualcommOptionsHtpPerformanceMode,
-          qualcomm_htp_performance_mode,
-          kLiteRtQualcommHtpPerformanceModeBurst, "HTP performance mode.");
+          qualcomm_htp_performance_mode, kLiteRtQualcommHtpPerformanceModeBurst,
+          "HTP performance mode.");
 
 ABSL_FLAG(std::vector<std::string>, qualcomm_dump_tensor_ids, {},
           "Debug Feature. Ids to dump as outputs. Comma-separated list of "
@@ -209,6 +209,12 @@ std::string AbslUnparseFlag(LiteRtQualcommOptionsProfiling options) {
       return "linting";
   }
 }
+
+ABSL_FLAG(
+    std::string, qualcomm_ir_json_dir, "",
+    "Qnn IR JSON directory. If provided, you can obtain Qnn IR in Qnn JSON "
+    "format.");
+
 // NOLINTEND(*alien-types*)
 
 namespace litert::qualcomm {
@@ -244,6 +250,9 @@ Expected<QualcommOptions> QualcommOptionsFromFlags() {
                   int32_ids.push_back(std::stoi(id));
                 });
   opts.SetDumpTensorIds(int32_ids);
+
+  const auto ir_json_dir = absl::GetFlag(FLAGS_qualcomm_ir_json_dir);
+  opts.SetIrJsonDir(ir_json_dir);
 
   return opts;
 }
--- a/litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
+++ b/litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
@@ -26,6 +26,7 @@
 
 #include "absl/container/flat_hash_map.h"  // from @com_google_absl
 #include "absl/container/flat_hash_set.h"  // from @com_google_absl
+#include "absl/strings/str_cat.h"  // from @com_google_absl
 #include "absl/strings/string_view.h"  // from @com_google_absl
 #include "absl/types/span.h"  // from @com_google_absl
 #include "litert/c/litert_common.h"
@@ -83,6 +84,7 @@
 #include "litert/vendors/qualcomm/core/builders/transpose_op_builder.h"
 #include "litert/vendors/qualcomm/core/builders/unpack_op_builder.h"
 #include "litert/vendors/qualcomm/core/common.h"
+#include "litert/vendors/qualcomm/core/dump/dump_graph.h"
 #include "litert/vendors/qualcomm/core/transformation/graph_to_graph.h"
 #include "litert/vendors/qualcomm/core/utils/miscs.h"
 #include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
@@ -1063,6 +1065,18 @@ LiteRtStatus MapGraph(QnnManager& qnn, Qnn_ContextHandle_t context_handle,
     LITERT_RETURN_IF_ERROR(ConvertOp(options.GetUseHtpPreference(), op,
                                      tensor_pool, input_tensors, output_tensors,
                                      op_wrappers));
+
+    if (op.Outputs().size() > 0 && !op.Outputs()[0].Name().empty()) {
+      // Add op namespace based on the first output tensor name.
+      if (size_t pos = op.Outputs()[0].Name().find(";");
+          pos != std::string_view::npos) {
+        auto op_namespace = op.Outputs()[0].Name().substr(0, pos);
+        for (auto& op_wrapper : op_wrappers) {
+          op_wrapper.AddPrefixToName(absl::StrCat(op_namespace, "/"));
+        }
+      }
+    }
+    // Move op_wrappers to graph_op_wrappers.
     std::move(op_wrappers.begin(), op_wrappers.end(),
               std::back_inserter(graph_op_wrappers));
   }
@@ -1083,6 +1097,12 @@ LiteRtStatus MapGraph(QnnManager& qnn, Qnn_ContextHandle_t context_handle,
                      options.GetUseQint16AsQuint16());
     }
     qnn.Api()->graphAddNode(graph_mapper.QnnGraph(), op_wrapper.GetOpConfig());
+  }
+
+  // Dump IR Json to understand Qnn graph.
+  if (!options.GetIrJsonDir().empty()) {
+    ::qnn::DumpIrJson(created_tensors, graph_op_wrappers,
+                      options.GetIrJsonDir(), qnn_graph_name);
   }
 
   LITERT_RETURN_STATUS_IF_QNN_NOT_OK(graph_mapper.Finalize());
--- a/litert/vendors/qualcomm/core/builders/op_builder.cc
+++ b/litert/vendors/qualcomm/core/builders/op_builder.cc
@@ -214,9 +214,9 @@ OpWrapper& CreateOpWrapper(std::vector<OpWrapper>& ops, const char* op_type) {
           {QNN_OP_UN_PACK, QnnOpCode::kUnPack},
       };
 
-  const auto op_count = ops.size();
-  const auto name = "op_type_" + std::string(op_type) + "_op_count_" +
-                    std::to_string(op_count);
+  // TODO(jiunkaiy): Remove the static op_index to ensure each op has a unique name.
+  static uint32_t op_index = 0;
+  const auto name = std::string(op_type) + "_" + std::to_string(op_index++);
 
   return ops.emplace_back(std::move(name), op_type, code_type_map->at(op_type));
 }
--- a/litert/vendors/qualcomm/core/common.cc
+++ b/litert/vendors/qualcomm/core/common.cc
@@ -58,6 +58,12 @@ std::vector<std::int32_t> Options::GetDumpTensorIds() const {
   return dump_tensor_ids_;
 }
 
+const absl::string_view Options::GetIrJsonDir() const { return ir_json_dir_; }
+
+void Options::SetIrJsonDir(absl::string_view ir_json_dir) {
+  ir_json_dir_ = ir_json_dir;
+}
+
 std::string Options::Dump() const {
   static constexpr absl::string_view kQnnOptionsDumpFormat =
       "\
@@ -68,14 +74,15 @@ UseHtpPreference: %v\n\
 UseQint16AsQuint16: %v\n\
 EnableWeightSharing: %v\n\
 HtpPerformanceMode: %d\n\
-DumpTensorIds: %s\n";  // NOLINT
+DumpTensorIds: %s\n\
+IrJsonDir: %s\n";  // NOLINT
 
   std::string dump_tensor_ids = absl::StrJoin(dump_tensor_ids_, ",");
 
   return absl::StrFormat(kQnnOptionsDumpFormat, log_level_, profiling_,
                          use_htp_preference_, use_qint16_as_quint16_,
                          enable_weight_sharing_, htp_performance_mode_,
-                         dump_tensor_ids);
+                         dump_tensor_ids, ir_json_dir_);
 }
 
 }  // namespace qnn
--- a/litert/vendors/qualcomm/core/common.h
+++ b/litert/vendors/qualcomm/core/common.h
@@ -8,6 +8,8 @@
 #include <string>
 #include <vector>
 
+#include "absl/strings/string_view.h"  // from @com_google_absl
+
 // c++ enum and wrapper without dependency.
 namespace qnn {
 
@@ -61,6 +63,9 @@ class Options {
   void SetDumpTensorIds(const std::vector<std::int32_t>& ids);
   std::vector<std::int32_t> GetDumpTensorIds() const;
 
+  const absl::string_view GetIrJsonDir() const;
+  void SetIrJsonDir(absl::string_view ir_json_dir);
+
   std::string Dump() const;
 
  private:
@@ -71,6 +76,7 @@ class Options {
   bool enable_weight_sharing_ = false;
   HtpPerformanceMode htp_performance_mode_ = HtpPerformanceMode::kDefault;
   std::vector<std::int32_t> dump_tensor_ids_;
+  std::string ir_json_dir_;
 };
 
 }  // namespace qnn
--- a/litert/vendors/qualcomm/core/dump/dump_graph.cc
+++ b/litert/vendors/qualcomm/core/dump/dump_graph.cc
@@ -1,0 +1,317 @@
+// Copyright (c) Qualcomm Innovation Center, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <filesystem>
+#include <fstream>
+#include <string>
+#include <string_view>
+#include <unordered_set>
+
+#include "absl/container/flat_hash_set.h"  // from @com_google_absl
+#include "absl/types/span.h"  // from @com_google_absl
+#include "litert/vendors/qualcomm/core/utils/log.h"
+#include "litert/vendors/qualcomm/core/utils/miscs.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "nlohmann/json.hpp"
+#include "QnnTypes.h"  // from @qairt
+
+namespace qnn {
+
+namespace {
+template <typename T>
+nlohmann::json ReshapeDataRecursive(uint32_t cur_index, T* data,
+                                    uint32_t num_elements,
+                                    absl::Span<uint32_t> dims) {
+  if (dims.empty()) {
+    return nlohmann::json();
+  }
+
+  uint32_t size = dims[0];
+  nlohmann::json nested_array = nlohmann::json::array();
+  if (dims.size() == 1) {
+    for (uint32_t i = 0; i < size; ++i) {
+      if (cur_index < num_elements) {
+        nested_array.emplace_back(data[cur_index++]);
+      } else {
+        QNN_LOG_ERROR("The data size for tensor_params does not match.");
+        // Fill with 0 if array is smaller than the specified dimensions.
+        nested_array.emplace_back(0);
+      }
+    }
+  } else {
+    absl::Span<uint32_t> sub_dims = dims.subspan(1);
+    for (uint32_t i = 0; i < size; ++i) {
+      nested_array.emplace_back(
+          ReshapeDataRecursive<T>(cur_index, data, num_elements, sub_dims));
+    }
+  }
+  return nested_array;
+}
+
+template <typename T>
+nlohmann::json ReshapeData(void* buffer_data, uint32_t buffer_size,
+                           absl::Span<uint32_t> dims) {
+  T* data = static_cast<T*>(buffer_data);
+  uint32_t ind = 0;
+  return ReshapeDataRecursive<T>(ind, data, buffer_size / sizeof(T), dims);
+}
+}  // namespace
+
+nlohmann::json SerializeQuantParamToJson(
+    const Qnn_QuantizeParams_t& quant_params) {
+  nlohmann::json qnn_quant_params = {
+      {"definition", quant_params.encodingDefinition},
+      {"encoding", quant_params.quantizationEncoding}};
+  // TODO (jiunkaiy): Support more quant encoding.
+  if (quant_params.quantizationEncoding ==
+      QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+    qnn_quant_params["scale_offset"] = {
+        {"scale", quant_params.scaleOffsetEncoding.scale},
+        {"offset", quant_params.scaleOffsetEncoding.offset}};
+  } else if (quant_params.quantizationEncoding ==
+             QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
+    qnn_quant_params["scale_offset"] = {
+        {"bitwidth", quant_params.bwScaleOffsetEncoding.bitwidth},
+        {"scale", quant_params.bwScaleOffsetEncoding.scale},
+        {"offset", quant_params.bwScaleOffsetEncoding.offset}};
+  } else if (quant_params.quantizationEncoding ==
+             QNN_QUANTIZATION_ENCODING_AXIS_SCALE_OFFSET) {
+    std::vector<nlohmann::json> scale_offsets;
+    uint32_t num_scale_offsets =
+        quant_params.axisScaleOffsetEncoding.numScaleOffsets;
+    scale_offsets.reserve(num_scale_offsets);
+    for (int i = 0; i < num_scale_offsets; ++i) {
+      scale_offsets.emplace_back(nlohmann::json{
+          {"scale", quant_params.axisScaleOffsetEncoding.scaleOffset[i].scale},
+          {"offset",
+           quant_params.axisScaleOffsetEncoding.scaleOffset[i].offset}});
+    }
+    qnn_quant_params["axis_scale_offset"] = {
+        {"axis", quant_params.axisScaleOffsetEncoding.axis},
+        {"num_scale_offsets", num_scale_offsets},
+        {"scale_offsets", scale_offsets},
+    };
+  } else if (quant_params.quantizationEncoding ==
+             QNN_QUANTIZATION_ENCODING_BW_AXIS_SCALE_OFFSET) {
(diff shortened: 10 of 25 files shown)


litert/c/options/litert_qualcomm_options.cc
```diff
@@ -244,3 +246,25 @@ LiteRtStatus LiteRtQualcommOptionsGetProfiling(
 
   return kLiteRtStatusOk;
 }
+
+LiteRtStatus LiteRtQualcommOptionsSetQnnJsonPath(LiteRtQualcommOptions options,
```
ghost: Let's add some test for flags.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -119,6 +119,12 @@ class TensorWrapper final {
 
   Qnn_Tensor_t& GetQnnTensor() { return qnn_tensor_; }
 
+  const uint32_t GetQnnTensorId() const { return qnn_tensor_.v1.id; }
+
+  const Qnn_Tensor_t& GetQnnTensor() const { return qnn_tensor_; }
+
+  const char* GetName() const { return qnn_tensor_.v1.name; }
+
```
ghost: Also, add some nice tests here :)

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -198,5 +198,14 @@ TEST(OpWrapperTest, SwapOutputsTest) {
...
 
+TEST(OpWrapperTest, ChangeOpName) {
+  OpWrapper op_wrapper_1{"name", "OP_TYPE", QnnOpCode::kUnknown};
+  EXPECT_STREQ(op_wrapper_1.GetOpConfig().v1.name, "name");
+  op_wrapper_1.AddNamespace("namespace");
+  EXPECT_STREQ(op_wrapper_1.GetOpConfig().v1.name, "namespace/name");
+  op_wrapper_1.AppendName("new");
+  EXPECT_STREQ(op_wrapper_1.GetOpConfig().v1.name, "namespace/name_new");
+}
+
```
ghost: Great!

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
```diff
@@ -156,4 +158,12 @@ std::vector<std::reference_wrapper<TensorWrapper>> OpWrapper::GetAllTensors() {
   return ret;
 };
 
+void OpWrapper::AddNamespace(absl::string_view op_namespace) {
+  name_ = absl::StrCat(op_namespace, "/", name_);
+}
+
+void OpWrapper::AppendName(absl::string_view name) {
```
ghost: NIT:

```suggestion
void OpWrapper::AppendNameWithUnderscore(absl::string_view name) {
```

litert/cc/options/litert_qualcomm_options.cc
```diff
@@ -136,6 +136,18 @@ std::vector<std::int32_t> QualcommOptions::GetDumpTensorIds() {
   return dump_ids;
 }
 
+void QualcommOptions::SetQnnJsonPath(const char* qnn_json_path) {
```
ghost:
```suggestion
void QualcommOptions::SetQnnJsonPath(const std::string& qnn_json_path) {
```

litert/cc/options/litert_qualcomm_options.cc
```diff
@@ -136,6 +136,18 @@ std::vector<std::int32_t> QualcommOptions::GetDumpTensorIds() {
...
+  internal::AssertOk(LiteRtQualcommOptionsSetQnnJsonPath, Data(),
+                     qnn_json_path);
```
ghost:
```suggestion
                     qnn_json_path.c_str());
```

litert/tools/flags/vendors/qualcomm_flags.cc
```diff
@@ -245,6 +250,9 @@ Expected<QualcommOptions> QualcommOptionsFromFlags() {
                 });
   opts.SetDumpTensorIds(int32_ids);
 
+  const auto qnn_json_path = absl::GetFlag(FLAGS_qualcomm_qnn_json_path);
+  opts.SetQnnJsonPath(qnn_json_path.c_str());
```
ghost:
```suggestion
  opts.SetQnnJsonPath(qnn_json_path);
```
Pass `const string&` here in c++ layer.

litert/vendors/qualcomm/core/common.cc
```diff
@@ -4,6 +4,7 @@
 #include "litert/vendors/qualcomm/core/common.h"
 
 #include <cstdint>
+
```
ghost:
```suggestion
```

litert/cc/options/litert_qualcomm_options.h
```diff
@@ -61,6 +61,9 @@ class QualcommOptions : public OpaqueOptions {
   void SetDumpTensorIds(const std::vector<std::int32_t>& ids);
   std::vector<std::int32_t> GetDumpTensorIds();
 
+  void SetQnnJsonPath(const char* profiling);
+  const char* GetQnnJsonPath();
```
ghost: In C++ layer, can we also return string here?
Then `move` to the member of `class Options` in `common.h`.

litert/vendors/qualcomm/core/common.cc
```diff
@@ -58,6 +59,14 @@ std::vector<std::int32_t> Options::GetDumpTensorIds() const {
...
 
+const absl::string_view Options::GetQnnJsonPath() const {
+  return qnn_json_path_;
+}
+
+void Options::SetQnnJsonPath(const char* qnn_json_path) {
+  qnn_json_path_ = qnn_json_path;
+}
+
 std::string Options::Dump() const {
```
ghost, line 67: Would you also like to dump the json path here?

litert/vendors/qualcomm/core/common.h
```diff
@@ -61,6 +63,9 @@ class Options {
   void SetDumpTensorIds(const std::vector<std::int32_t>& ids);
   std::vector<std::int32_t> GetDumpTensorIds() const;
 
+  const absl::string_view GetQnnJsonPath() const;
+  void SetQnnJsonPath(const char* qnn_json_path);
```
ghost: As mentioned above, would passing a `std::string` be better?

litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
```diff
@@ -928,6 +929,18 @@ LiteRtStatus MapGraph(QnnManager& qnn, Qnn_ContextHandle_t context_handle,
...
+    // TODO(jiunkaiy): Change to the op namespace if the literal op contains
+    // this information.
+    if (op.Outputs().size() > 0 && !op.Outputs()[0].Name().empty()) {
+      // Add op namespace based on the first output tensor name.
+      for (auto& op_wrapper : op_wrappers) {
+        size_t pos = op.Outputs()[0].Name().find(";");
+        if (pos != std::string_view::npos) {
+          op_wrapper.AddNamespace(op.Outputs()[0].Name().substr(0, pos));
+        }
+      }
```
ghost:
```suggestion
      size_t pos = op.Outputs()[0].Name().find(";");
      if (pos != std::string_view::npos) {
          auto namespace = op.Outputs()[0].Name().substr(0, pos);
          for (auto& op_wrapper : op_wrappers) {
              op_wrapper.AddNamespace(namespace);
          }
      }
```

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+
+  // Dump Qnn Ops.
+  std::unordered_set<uint32_t> param_tensor_ids;
+  std::unordered_set<std::string> op_types;
+  for (auto& op : graph_op_wrappers) {
+    const Qnn_OpConfig_t op_config = op.GetOpConfig();
+    nlohmann::json qnn_node_json = nlohmann::json::object();
+    // Create input_names and output_names.
+    qnn_node_json["input_names"] = nlohmann::json::array();
+    for (uint32_t i = 0; i < op_config.v1.numOfInputs; ++i) {
```
ghost: Will it cause issue of op version becomes v2 in the future? Not sure if getting attr from OpWrapper would be a better idea.

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+    };
+  } else {
+    QNN_LOG_WARNING(
+        "Quantization encoding: %u is not supported in Qnn "
+        "Json dump",
+        quant_params.quantizationEncoding);
+  }
+}
+
+void AddScalarParams(const Qnn_Param_t& params, nlohmann::json& qnn_node_json) {
```
ghost: How about passing `qnn_node_json["scalar_params"]` instead of the whole json.

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+#include "litert/vendors/qualcomm/core/utils/log.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "nlohmann/json.hpp"
+#include "QnnTypes.h"  // from @qairt
+
+namespace qnn {
+
+namespace {
+void AddQnnTensor(const Qnn_TensorV1_t& qnn_tensor,
```
ghost: Can discuss offline how to handle V2 tensor

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+    }
+  }
+  return nested_array;
+}
+
+nlohmann::json GetData(Qnn_DataType_t datatype,
+                       const Qnn_ClientBuffer_t& buffer,
+                       absl::Span<uint32_t> dims) {
+  uint32_t index = 0;
+  if (datatype == QNN_DATATYPE_INT_8 ||
```
ghost: How about using switch-case like above

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+  }
+}
+
+template <typename T>
+nlohmann::json ReshapeData(absl::Span<T> data, absl::Span<uint32_t> dims,
+                           uint32_t& index) {
+  if (dims.empty()) {
+    return nlohmann::json();
+  }
+
```
ghost: Check data size with dims

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+  uint32_t size = dims[0];
+  nlohmann::json nested_array = nlohmann::json::array();
+  if (dims.size() == 1) {
+    for (uint32_t i = 0; i < size; ++i) {
+      if (index < data.size()) {
+        nested_array.emplace_back(data[index++]);
+      } else {
+        // Fill with 0 if array is smaller than the specified dimensions.
```
ghost: Maybe we can consider remove this

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+  }
+  // Dump Qnn Tensors.
+  for (const TensorWrapper* tensor : tensor_wrappers) {
+    // Skip tensor params.
+    if (param_tensor_ids.count(tensor->GetQnnTensorId()) > 0) {
+      continue;
+    }
+    // Create tensors.
+    nlohmann::json qnn_tensor_json = nlohmann::json::object();
+    AddQnnTensor(tensor->GetQnnTensor().v1, qnn_tensor_json);
```
ghost: Can discuss offline about v2

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,325 @@
...
+    qnn_json["graph"]["tensors"][tensor->GetName()] = qnn_tensor_json;
+  }
+  // Dumpe Qnn op types.
+  qnn_json["op_types"] = op_types;
```
ghost: Can we leave some comment that the json only stores data of tensor params, not every tensor data.

litert/c/options/litert_qualcomm_options.cc
jiunkaiy: Added litert option test for LiteRtQualcommOptionsSetQnnJsonPath. Thanks!

litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
jiunkaiy: Got your point. Thank you!
I'll move the pos and op namespaces outside of the for loop. However, I won't use `namespace` as a variable name, as I don't think that's appropriate.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: In QNN design, they keep backward compatibility for structs, so v2 should only add new members at the end of the struct.

I retrieve op information from `op_config` instead of the litert `OpWrapper` intentionally, as the QNN JSON should directly reflect what is entered in QNN. This way, we can make sure all information stays consistent.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: The current QNN JSON (in ORT/Converter) does not contain the new members in Tensor V2, so we can read Tensor as Tensor V1 without issues.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: The current QNN JSON (in ORT/Converter) does not contain the new members in Tensor V2, so we can read Tensor as Tensor V1 without issues.

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
jiunkaiy: Thanks!

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Changed to switch-case and refactor `ReshapeData` function. Thank you!

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: I added `QNN_LOG_ERROR` to inform the user that something went wrong.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: I would like to keep this as it is, but include an error log.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Added some comment. Thank you!

litert/vendors/qualcomm/core/common.cc
jiunkaiy: Umm, I think printing the QNN JSON Path before the QNN JSON dump should be sufficient.

litert/cc/options/litert_qualcomm_options.h
jiunkaiy: After offline discussion, we retain const char* here.

litert/vendors/qualcomm/core/common.h
jiunkaiy: After offline discussion, we retain const char* here.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
jiunkaiy: Added tests for these three functions. Thank you!

litert/cc/options/litert_qualcomm_options.h
```diff
@@ -61,6 +61,9 @@ class QualcommOptions : public OpaqueOptions {
   void SetDumpTensorIds(const std::vector<std::int32_t>& ids);
   std::vector<std::int32_t> GetDumpTensorIds();
 
+  void SetQnnJsonDir(const std::string& profiling);
+  const char* GetQnnJsonDir();
```
weilhuan-quic: We can return `std::string_view` here because we guarantee `qnn_json_dir` is null terminated.

litert/c/options/litert_qualcomm_options_test.cc
```diff
@@ -153,6 +153,24 @@ TEST(LiteRtQualcommOptionsTest, Profiling) {
...
+      LiteRtQualcommOptionsSetQnnJsonDir(qualcomm_options, "tmp/"));
+
+  const char* qnn_json_dir;
+  LITERT_ASSERT_OK(
+      LiteRtQualcommOptionsGetQnnJsonDir(qualcomm_options, &qnn_json_dir));
+  EXPECT_STREQ(qnn_json_dir, "tmp/");
+
+  LiteRtDestroyOpaqueOptions(options);
+}
+
```
weilhuan-quic, line 172: We had better also test c++ api in `TEST(QualcommOptionsTest, CppApi)`

litert/c/options/litert_qualcomm_options.cc
```diff
@@ -35,6 +36,7 @@ struct LiteRtQualcommOptionsT {
   LiteRtQualcommOptionsHtpPerformanceMode htp_performance_mode =
       kLiteRtQualcommHtpPerformanceModeDefault;
   std::vector<std::int32_t> dump_tensor_ids;
+  std::string qnn_json_dir;
```
weilhuan-quic: Maybe we can call it qnn ir json.
```suggestion
  std::string qnn_ir_json_dir;
```

litert/vendors/qualcomm/core/builders/op_builder.cc
```diff
@@ -214,9 +214,8 @@ OpWrapper& CreateOpWrapper(std::vector<OpWrapper>& ops, const char* op_type) {
           {QNN_OP_UN_PACK, QnnOpCode::kUnPack},
       };
 
-  const auto op_count = ops.size();
-  const auto name = "op_type_" + std::string(op_type) + "_op_count_" +
-                    std::to_string(op_count);
+  static uint32_t op_index = 0;
```
weilhuan-quic, line 218: Can you add a TODO for me to remove the static counter? is this for unique op name?

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -119,6 +119,12 @@ class TensorWrapper final {
 
   Qnn_Tensor_t& GetQnnTensor() { return qnn_tensor_; }
 
+  const uint32_t GetQnnTensorId() const { return qnn_tensor_.v1.id; }
+
+  const Qnn_Tensor_t& GetQnnTensor() const { return qnn_tensor_; }
+
+  const char* GetName() const { return qnn_tensor_.v1.name; }
```
weilhuan-quic: Create a new `std::string` and return `std::string` might be safer than return the raw point in `qnn_tensor_`.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -119,6 +119,12 @@ class TensorWrapper final {
 
   Qnn_Tensor_t& GetQnnTensor() { return qnn_tensor_; }
 
+  const uint32_t GetQnnTensorId() const { return qnn_tensor_.v1.id; }
```
weilhuan-quic:
```suggestion
  uint32_t GetId() const { return qnn_tensor_.v1.id; }
```

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
```diff
@@ -156,4 +158,12 @@ std::vector<std::reference_wrapper<TensorWrapper>> OpWrapper::GetAllTensors() {
...
 };
 
+void OpWrapper::AddNamespace(absl::string_view op_namespace) {
+  name_ = absl::StrCat(op_namespace, "/", name_);
+}
+
+void OpWrapper::AppendNameWithUnderscore(absl::string_view name) {
+  name_ = absl::StrCat(name_, "_", name);
+}
+
```
weilhuan-quic: Because `OpWrapper` is designed to wrap `Qnn_OpConfig_t`, and namespace is not related to  `Qnn_OpConfig_t`. So I prefer passing a proper name in the constructor of `OpWrapper` instead of modify the name later. Can discuss offline.

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,324 @@
...
+}  // namespace
+
+void DumpQnnJson(
+    const absl::flat_hash_set<const TensorWrapper*>& tensor_wrappers,
+    std::vector<OpWrapper>& graph_op_wrappers, std::string_view json_dir,
+    std::string_view graph_name) {
+  CreateDirectoryRecursive(json_dir);
+  std::string qnn_json_path =
+      (json_dir.back() == '/') ? json_dir.data() : absl::StrCat(json_dir, "/");
+  qnn_json_path = absl::StrCat(qnn_json_path, graph_name, ".json");
```
weilhuan-quic: Nit: use `std::filesystem::path`

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,324 @@
...
+    qnn_node_json["macs_per_inference"] = "";
+    qnn_node_json["type"] = op_config.v1.typeName;
+    // Record seen op type in a set.
+    op_types.emplace(op_config.v1.typeName);
+    // Create scalar_params and tensor_params.
+    qnn_node_json["scalar_params"] = nlohmann::json::object();
+    qnn_node_json["tensor_params"] = nlohmann::json::object();
+    for (uint32_t i = 0; i < op_config.v1.numOfParams; ++i) {
+      if (op_config.v1.params[i].paramType == QNN_PARAMTYPE_SCALAR) {
+        AddScalarParams(op_config.v1.params[i], qnn_node_json["scalar_params"]);
```
weilhuan-quic: Nit: we can pass `qnn_node_json` and let `AddScalarParams` handle "scalar_params" inside.

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,324 @@
...
+      nested_array.emplace_back(
+          ReshapeData<T>(index, buffer_data, data_size, sub_dims));
+    }
+  }
+  return nested_array;
+}
+
+nlohmann::json GetData(Qnn_DataType_t datatype,
+                       const Qnn_ClientBuffer_t& buffer,
+                       absl::Span<uint32_t> dims) {
```
weilhuan-quic: We can access what we need in `Qnn_Tensor_t`, caller won't need to pass so many param. 
```suggestion
nlohmann::json ConvertTensorData(const Qnn_Tensor_t &) {
```

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,324 @@
...
+  uint32_t size = buffer.dataSize;
+  uint32_t ind = 0;
+  void* data = buffer.data;
+  switch (datatype) {
+    case QNN_DATATYPE_INT_8:
+    case QNN_DATATYPE_SFIXED_POINT_8:
+      return ReshapeData<int8_t>(ind, data, size / sizeof(int8_t), dims);
```
weilhuan-quic: `sizeof(T)` can be inferenced in `ReshapeData`, caller won't need to calculate it here. Just provide data pointer and the data size. 

BTW, `data_size` is not proper in `ReshapeData`, actually it is element size.

weilhuan-quic: `ind` must be 0 here, we can have a `ReshapeData` as entry point where handle the index and the element size.

e.g.
```cpp
ReshapeData<T>(void *data, size_t  data_size, dims)` 
-> ReshapeDataRecursive<T>(size_t cur_index, T* data, size_t element_size, dims) 
```

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,324 @@
...
+  }
+  // Dump Qnn Tensors.
+  for (const TensorWrapper* tensor : tensor_wrappers) {
+    // Skip tensor params.
+    if (param_tensor_ids.count(tensor->GetQnnTensorId()) > 0) {
+      continue;
+    }
+    // Create tensors.
+    nlohmann::json qnn_tensor_json = nlohmann::json::object();
+    AddQnnTensor(tensor->GetQnnTensor().v1, qnn_tensor_json);
```
weilhuan-quic: Do we need to dump static data for the static tensor here?

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,324 @@
...
+#include "litert/vendors/qualcomm/core/utils/miscs.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "nlohmann/json.hpp"
+#include "QnnTypes.h"  // from @qairt
+
+namespace qnn {
+
+namespace {
+void AddQnnTensor(const Qnn_TensorV1_t& qnn_tensor,
```
weilhuan-quic: I think we can have these convert function for test qnn types, just like what we do in wrappers. Then you can write unit tests for these function, instead of one unit test for a complete Qnn graph/op. Can offline discuss,

```
nlohmann::json SerializeTensorToJson(const Qnn_Tensor_t& qnn_tensor);
nlohmann::json SerializeOpToJson(const Qnn_OpConfig_t& qnn_op);
nlohmann::json SerializeScalarParamToJson(const Qnn_Scalar_t& qnn_scalar);
nlohmann::json SerializeTensorParamToJson(const Qnn_Tensor_t& qnn_tensor);
nlohmann::json SerializeQuantParamToJson(..);
...
```

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
jiunkaiy: Nice catch!

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Thanks!

litert/cc/options/litert_qualcomm_options.h
jiunkaiy: Thanks!

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: No, only tensor params. I have added comments based on @chunhsue previous suggestion.
```
  // Dump Qnn Ops.
  // Note that only the static data of tensorParam is stored in QNN Json.
```
image

Ref.:
https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-50/converters.html

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
jiunkaiy: Changed to std::string, thanks!

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Thanks, I have changed to `GetTensorData` with less param.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Thanks!

litert/c/options/litert_qualcomm_options.cc
jiunkaiy: I updated it to `ir_json_dir` to be consistent with QNN IrBackend. Thanks!

litert/c/options/litert_qualcomm_options_test.cc
jiunkaiy-qti: Added. Thanks!

jiunkaiy: Added. Thanks!

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
jiunkaiy: Change to
AddPrefixToName
AddSuffixToName

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Change to Serializexxx and added a few tests. Thanks!

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,319 @@
...
+    case QNN_DATATYPE_FLOAT_32:
+      return ReshapeData<float>(data, size, dims);
+    case QNN_DATATYPE_FLOAT_64:
+      return ReshapeData<double>(data, size, dims);
+    case QNN_DATATYPE_INT_64:
+      return ReshapeData<int64_t>(data, size, dims);
+    case QNN_DATATYPE_UINT_64:
+      return ReshapeData<uint64_t>(data, size, dims);
+    default:
+      QNN_LOG_ERROR(
```
chuntl-qti, line 213: Nit:
If the datatype is not supported, then the error log will show in the terminal for the json generation. I'm thinking if we should still add a return here, to return a json with "unsupported" message, then even the datatype is unsupported, it will be clear in the issued json file.
Btw, I agree that we can only rely on the error log to let developer know the unsupported issue.

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,319 @@
...
+      {"model.bin", "N/A"},
+      {"converter_command", ""},
+      {"copyright_str",
+       "Copyright (c) Qualcomm Innovation Center, Inc. All Rights Reserved."},
+      {"op_types", nlohmann::json::array()},
+      {"Total parameters", ""},
+      {"Total MACs per inference", ""},
+      {"graph",
+       {{"tensors", nlohmann::json::object()},
+        {"nodes", nlohmann::json::object()}}}};
```
chuntl-qti, line 276: Nit:

Do we want to add some information like timestamp, qnn version or anything else?

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,319 @@
...
+  }
+  return nested_array;
+}
+
+template <typename T>
+nlohmann::json ReshapeData(void* buffer_data, uint32_t buffer_size,
+                           absl::Span<uint32_t> dims) {
+  T* data = static_cast<T*>(buffer_data);
+  uint32_t ind = 0;
+  return ReshapeDataRecursive<T>(ind, data, buffer_size / sizeof(T), dims);
```
chuntl-qti, line 57: If the weight is huge, the json file will be huge too, not sure if we need to add some check here to prevent the json become very big.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: I would prefer to keep the current design since the QNN IR JSON is defined by the QNN converter. Adding an "unsupported" message to the JSON might cause issues when uploading it to Netron or other visualization tools.

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Since we need to follow QNN Model JSON format, we cannot add timestamp or qnn version I think 🤔 
https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-50/converters.html

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: We add tensor data only for tensor parameters (following QNN Model JSON format), which is relative small.
Static tensors, such as CONV2D weights, are not included.

litert/vendors/qualcomm/compiler/qnn_compose_graph.cc
```diff
@@ -1063,6 +1064,19 @@ LiteRtStatus MapGraph(QnnManager& qnn, Qnn_ContextHandle_t context_handle,
     LITERT_RETURN_IF_ERROR(ConvertOp(options.GetUseHtpPreference(), op,
                                      tensor_pool, input_tensors, output_tensors,
                                      op_wrappers));
+    // TODO(jiunkaiy): Change to the op namespace if the literal op contains
+    // this information.
+    if (op.Outputs().size() > 0 && !op.Outputs()[0].Name().empty()) {
+      // Add op namespace based on the first output tensor name.
+      size_t pos = op.Outputs()[0].Name().find(";");
+      if (pos != std::string_view::npos) {
```
ghost: NIT
```suggestion
      if (size_t pos = op.Outputs()[0].Name().find(";"); pos != std::string_view::npos) {
```

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -327,7 +327,7 @@ TEST(TensorWrapperTest, QnnTensorPerTensorQuantConstructTest) {
                                {1, 1, 3}};
   const auto& qnn_tensor = tensor_wrapper.GetQnnTensor();
   TensorWrapper tensor_wrapper_1(qnn_tensor);
-  Qnn_Tensor_t& ref = tensor_wrapper.GetQnnTensor();
+  Qnn_Tensor_t& ref = tensor_wrapper_1.GetQnnTensor();
```
ghost, line 330: Thanks for finding out the typo!

litert/vendors/qualcomm/core/wrappers/op_wrapper.h
```diff
@@ -63,6 +64,10 @@ class OpWrapper final {
 
   void ClearTensorParams();
 
+  void AddPrefixToName(absl::string_view op_namespace);
+
+  void AddSuffixToName(absl::string_view name);
```
weilhuan-quic: Nit:
```suggestion
  void AddPrefixToName(absl::string_view prefix);

  void AddSuffixToName(absl::string_view suffix);
```

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
```diff
@@ -156,4 +158,12 @@ std::vector<std::reference_wrapper<TensorWrapper>> OpWrapper::GetAllTensors() {
   return ret;
 };
 
+void OpWrapper::AddPrefixToName(absl::string_view op_namespace) {
+  name_ = absl::StrCat(op_namespace, "/", name_);
```
weilhuan-quic: I think "/" should be also provide in `op_namespace`.

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,319 @@
...
+  qnn_node_json["output_names"] = nlohmann::json::array();
+  for (uint32_t i = 0; i < op_config.v1.numOfOutputs; ++i) {
+    qnn_node_json["output_names"].emplace_back(
+        op_config.v1.outputTensors[i].v1.name);
+  }
+  // Create macs_per_inferences and op type.
+  qnn_node_json["macs_per_inference"] = "";
+  qnn_node_json["type"] = op_config.v1.typeName;
+  // Create scalar_params and tensor_params.
+  qnn_node_json["tensor_params"] = nlohmann::json::object();
```
weilhuan-quic, line 239: maybe we can also create `nlohmann::json::object()` for "scalar_params"?

litert/vendors/qualcomm/core/dump/dump_graph_test.cc
```diff
@@ -0,0 +1,220 @@
...
+#include "litert/vendors/qualcomm/core/tensor_pool.h"
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "nlohmann/json.hpp"
+
+namespace qnn {
+namespace {
+
+TEST(SerializeOpToJson, ElementWiseAdd) {
```
weilhuan-quic: Actually it is matmul op instead.
```suggestion
TEST(IrJsonDump, SerializeOpToJson) {
```

litert/vendors/qualcomm/core/dump/dump_graph_test.cc
```diff
@@ -0,0 +1,220 @@
...
+  EXPECT_EQ(qnn_op["output_names"][0], "2_qnn");
+  ASSERT_TRUE(qnn_op.contains("scalar_params"));
+  ASSERT_TRUE(qnn_op["scalar_params"].contains("transpose_in0"));
+  ASSERT_TRUE(qnn_op["scalar_params"].contains("transpose_in1"));
+  ASSERT_TRUE(qnn_op.contains("tensor_params"));
+  ASSERT_TRUE(qnn_op.contains("type"));
+  EXPECT_EQ(qnn_op["type"], "MatMul");
+}
+
+TEST(SerializeQuantParamToJson, RMSNorm_Axes) {
```
weilhuan-quic:
```suggestion
TEST(IrJsonDump, SerializeQuantParamToJson) {
```

litert/vendors/qualcomm/core/dump/dump_graph.cc
jiunkaiy: Changed to nlohmann::json::object() for "scalar_params". Thanks!

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
jiunkaiy: AddPrefixToName without adding "/". Thanks!

chuntl-qti (approved):

litert/vendors/qualcomm/core/dump/dump_graph.cc
```diff
@@ -0,0 +1,317 @@
...
+#include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
+#include "litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h"
+#include "nlohmann/json.hpp"
+#include "QnnTypes.h"  // from @qairt
+
+namespace qnn {
+
+namespace {
+template <typename T>
+nlohmann::json ReshapeDataRecursive(uint32_t cur_index, T* data,
```
weilhuan-quic, line 23: I think it should be reference of `cur_index`.
```suggestion
nlohmann::json ReshapeDataRecursive(uint32_t &cur_index, T* data,
```

ghost: That is right! Thank you!
Btw, I will move to new PR.

PR #2620, opened 2025-06-25
Add implementation for Intel OpenVINO compiler plugin

Rjasuja:


litert/vendors/intel_openvino/compiler/decoder.h
```diff
@@ -0,0 +1,148 @@
...
+#define ODML_LITERT_LITERT_VENDORS_OPENVINO_COMPILER_DECODER_H_
+
+#include <vector>
+#include <string>
+
+#include "openvino/frontend/tensorflow_lite/decoder.hpp"
+
+#include "litert/c/litert_logging.h"
+#include "litert/cc/litert_model.h"
+#include "openvino/frontend/tensorflow_lite/decoder.hpp"
```
MatthiasKreileder: Already included on line 10

litert/vendors/intel_openvino/compiler/decoder.h
```diff
@@ -0,0 +1,148 @@
...
+
+namespace litert {
+namespace openvino {
+
+class DecoderOperation
+    : public ov::frontend::tensorflow_lite::DecoderBaseOperation {
```
MatthiasKreileder, line 31: There's nothing wrong with your approach, though I was wondering if you would consider 'composition' instead of inheritance here?

From what I understand composition would work just as well here. This has the advantage that it allows you to make this class, which is part of LiterRT, consistent with the rest of the codebase + coding guidelines. E.g. you could address the comments from other reviewers about applying pascal case for methods etc).

Composition is also 'preferred' for Google C++ code https://google.github.io/styleguide/cppguide.html#Inheritance (though inheritance, especially public, isn't wrong either).

Let me know what you think. If you agree then we should apply this consistently also to other classes, like `DecoderTensor`, `GraphIteratorDelegate` etc.

litert/vendors/intel_openvino/compiler/decoder.cc
```diff
@@ -0,0 +1,672 @@
...
+      output_tensor_info_(output_tensor_info),
+      litert_op_(litert_op.Get()),
+      litert_op_code_(litert_op.Code()) {
+    op_type_ = GetOvOpType(litert_op_code_);
+  op_name_ = op_type_ + "_id_" + std::to_string(node_index);
+  LITERT_LOG(LITERT_VERBOSE, "op_type(%s) op_name(%s)", op_type_.c_str(),
+             op_name_.c_str());
+}
+
+#define DECODER_CHECK_STATUS(status, attr)                                \
```
MatthiasKreileder: Please consider using [`LITERT_RETURN_IF_ERROR`](https://github.com/google-ai-edge/LiteRT/blob/0348ffbe4232df35ab2651e6383528b3d8bf792f/litert/cc/litert_macros.h#L32) which has been carefully designed for this purpose (IIUC).

litert/vendors/intel_openvino/compiler/decoder.h
Rjasuja: Inheritance is helping achieve code re-usability from openvino tensorflow lite frontend code [here](https://github.com/openvinotoolkit/openvino/tree/master/src/frontends/tensorflow_lite). We would lose the benefit if we try to use composition here. It would mean to maintain node and tensor creation within the openvino plugin which is at present being re-used from the link above. We implemented it this way for easier maintainability and code re-usability.

Rjasuja: GraphIteratorDelegate, DecoderTensor and DecoderOperation are the classes for which we need inheritance such that  all the magic happens inside the frontend without the plugin's responsibility for maintenance.

Rjasuja:
@MatthiasKreileder Addressed all comments. Please review

MatthiasKreileder (approved):

ShikChen:
Hi, is there anything blocking this from being merged?

litert/vendors/intel_openvino/compiler/graph_iterator.h
```diff
@@ -0,0 +1,105 @@
...
+#include <openvino/frontend/tensorflow_lite/graph_iterator.hpp>
+#include <openvino/frontend/tensorflow_lite/quantization_info.hpp>
+#include <unordered_set>
+#include <vector>
+
+#include "decoder.h"
+#include "litert/c/litert_logging.h"
+#include "litert/c/litert_model.h"
+#include "litert/cc/litert_model.h"
+#include "litert/core/model/model.h"
```
YunanAZ: Use internal api in vendor plugin is disallowed, please remove corresponding usages of api defined in this file.

YunanAZ (changes requested):

PR #2670, opened 2025-07-01
remove gitmodules. workpsace by default is using http_archive

ecalubaquib:
remove gitmodules. workpsace by default is using http_archive pointing to master
through source_repo.


ecalubaquib:
the WORKSPACE by default is using http_archive
https://github.com/google-ai-edge/LiteRT/blob/main/WORKSPACE#L32
with that in mind we can remove the .gitmodules file

fcoUnda (approved):

PR #3018, opened 2025-07-28
Exclude linux header files for non-AHWB supported builds

Rjasuja:
Disable the include of unistd.h and sys/socket.h linux header files for enabling Windows builds

### DIFF
--- a/litert/vendors/intel_openvino/dispatch/device_context.cc
+++ b/litert/vendors/intel_openvino/dispatch/device_context.cc
@@ -19,8 +19,10 @@
 #endif  // __ANDROID__
 
 #include <string.h>
+#if LITERT_HAS_AHWB_SUPPORT
 #include <sys/socket.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


litert/vendors/intel_openvino/dispatch/device_context.cc
```diff
@@ -19,8 +19,10 @@
 #endif  // __ANDROID__
 
 #include <string.h>
+#if LITERT_HAS_AHWB_SUPPORT
```
MatthiasKreileder, line 22: There's actually a more natural way of guarding window's specific things, via `LITERT_WINDOWS_OS` from here https://github.com/google-ai-edge/LiteRT/blob/0295c3e3af11f85bc60adbb1dca405955a010f39/litert/cc/litert_macros.h#L94

Could you please include that header (and specify the target as a `dep`) and use that? Thanks

MatthiasKreileder (approved):

PR #3167, opened 2025-08-07
Support OpenVINO Compiler plugin and Dispatch API on Windows

fujunwei:
1, Generate LiteRtCompilerPlugin.dll and LiteRtDispatch.dll
2, Support kLiteRtTensorBufferTypeHostMemory TensorBuffer to upload input data to NPU


litert/vendors/c/litert_dispatch.h
```diff
@@ -219,74 +219,73 @@ typedef enum LiteRtDispatchNodeType {
...
                                         LiteRtDispatchNodeId node_id,
                                         const char* key, const char* value);
 
 // Add an annotation to a specified edge.
-LiteRtStatus LiteRtDispatchAnnotateEdge(LiteRtDispatchGraph* graph,
+LITERT_CAPI_EXPORT LiteRtStatus LiteRtDispatchAnnotateEdge(LiteRtDispatchGraph* graph,
                                         LiteRtDispatchEdgeId edge_id,
                                         const char* key, const char* value);
 
-// Create an invocation context from a given graph.
```
Rjasuja, line 288: Can you help put the comment back

litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -54,6 +54,40 @@ litert_dynamic_lib(
     }),
 )
 
+litert_bin(
```
Rjasuja, line 55: Can you help explain the reason here for having binaries for windows instead of dynamic libraries.

litert/vendors/intel_openvino/compiler/graph_iterator.h
```diff
@@ -87,7 +87,10 @@ class GraphIteratorDelegate
   /// If there is no query for specific sub-graph iterator shouldn't be created
   /// idx should be in range 0..get_subgraph_size()-1
   std::shared_ptr<ov::frontend::tensorflow_lite::GraphIterator> get_subgraph(
-      size_t idx) const override {};
+      size_t idx) const override {
+    LITERT_LOG(LITERT_ERROR, "get_subgraph not implemented");
```
Rjasuja, line 91: Looks like a style formatting issue here

fujunwei: It has already used `clang-format  -style=google`.

litert/vendors/intel_openvino/compiler/BUILD
fujunwei: `litert_dynamic_lib` only can output lib file not dll, i don't find the root cause of it, any insights?

jl45G: [Google employee here]

This is a known issue, it only produces .so on unix at this point, since it enforces unix naming, something like libLiteRt*.so.

We also lack support for windows export scripts and windows's own generation rule. 

But I also don't think litert_bin will produce dll, may I ask how did you workaround this issue? Internally we will also try to address the limitation as soon as possible

fujunwei: [linkshared=True](https://bazel.build/reference/be/c-cpp#cc_binary.linkshared) can create dll on windows,  [here](https://github.com/bazelbuild/bazel/tree/master/examples/windows/dll) is a sample code.

jl45G: Got it, thanks, then let's use this solution in the meantime.

fujunwei:
Thanks for review, any other comments?

litert/vendors/intel_openvino/compiler/decoder.h
```diff
@@ -142,11 +144,13 @@ class DecoderTensor : public ov::frontend::tensorflow_lite::DecoderBaseTensor {
   /// \brief No operation for tensor
   const std::string &get_op_type() const override {
     LITERT_LOG(LITERT_ERROR, "get_op_type not implemented");
+    return "not implemented";
   };
 
   /// \brief No operation name for tensor
   const std::string &get_op_name() const override {
```
jl45G, line 150: binding temporary string to const string& isn't a good practice

litert/vendors/intel_openvino/compiler/decoder.h
```diff
@@ -125,11 +125,13 @@ class DecoderTensor : public ov::frontend::tensorflow_lite::DecoderBaseTensor {
   /// \brief No attributes for tensor
   ov::Any get_attribute(const std::string &name) const override {
     LITERT_LOG(LITERT_ERROR, "get_attribute not implemented");
+    return nullptr;
```
jl45G, line 128: return ov::Any{}

jl45G: return ov::any{}

litert/c/litert_common.h
```diff
@@ -21,6 +21,18 @@
 extern "C" {
 #endif  // __cplusplus
 
+// Define LITERT_CAPI_EXPORT macro to export a function properly with a shared
+// library.
+#if defined(_WIN32)
```
jl45G: use litert macros here: https://github.com/google-ai-edge/LiteRT/blob/92b8d54810cb8113aa0cae49e5b824fcb7028307/litert/cc/litert_macros.h#L94

litert/c/litert_common.h
```diff
@@ -21,6 +21,18 @@
...
+#ifdef LITERT_COMPILE_LIBRARY
+#define LITERT_CAPI_EXPORT __declspec(dllexport)
+#else
+#define LITERT_CAPI_EXPORT
```
jl45G: iirc, on msvc, it's preferred to use __declspec(dllimport) when consuming the DLL for better codegen

so
#if defined(_WIN32)
  #ifdef LITERT_COMPILE_LIBRARY
    #define LITERT_CAPI_EXPORT __declspec(dllexport)
  #else
    #define LITERT_CAPI_EXPORT __declspec(dllimport)
  #endif
#else
  #define LITERT_CAPI_EXPORT __attribute__((visibility("default")))
#endif

jl45G (approved):
overall looks good. Please rebase your branch on HEAD before submission.

fujunwei:
Thanks for review, @LukeBoyer PTAL, thanks.

fujunwei:
@MatthiasKreileder i can't view the internal error messages, could you help me? thanks.

MatthiasKreileder:
>  @MatthiasKreileder i can't view the internal error messages, could you help me? thanks.

We need to also support clang for Windows for the Google-internal build and for the OSS build I suppose you and others will want to use MSVC.

In order to get around the internal errors I introduced a build_config. PTAL https://github.com/google-ai-edge/LiteRT/blob/2729579649de386ddd5f929c6f035523181c62fc/litert/runtime/dispatch/BUILD#L35 

Please let me know if this is problematic for you, we can find another approach then.

fujunwei:
The configuration can work on my side, thanks for merging the PR.

PR #3564, opened 2025-09-16
Qualcomm AI Engine Direct - Check build id in the debugger.

chunhsue-qti:
Summary:
- Check build id between context binary and current qairt sdk to avoid qairt feature mismatch.

# Test
```
RuntimeError: The context binary is compiled with BUILD ID: v2.41.0.251103115413_190331 != current qairt sdk BUILD ID: v2.39.0.250925215840_163802
```

### DIFF
--- a/litert/vendors/qualcomm/debugger/optrace_profiling/run.py
+++ b/litert/vendors/qualcomm/debugger/optrace_profiling/run.py
@@ -87,12 +87,10 @@ def _get_ctx_bin_info(ctx_bin_path: str, qairt_sdk: Path) -> dict[str, Any]:
   if version_parse(bin_id) == version_parse(qairt_id):
     logging.info("Build ID matches.")
   else:
-    logging.warning(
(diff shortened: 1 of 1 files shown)


litert/vendors/qualcomm/debugger/optrace_profiling/run.py
```diff
@@ -441,6 +441,10 @@ def _generate_ctx_bin(model_path: Path, soc_model: str) -> Path:
       default=default_qairt_sdk,
   )
   args = parser.parse_args()
+  
+  # LiteRT uses LITERT_QAIRT_SDK, align it with the one from args
+  os.environ["LITERT_QAIRT_SDK"] = f"{str(args.qairt_sdk.parent)}/"
```
weilhuan-qti: If user specify a qairt_sdk which is not inside `${LiteRtRoot}/third_party`, we may need further check instead of directly override `${LITERT_QAIRT_SDK}`.

weilhuan-qti (approved):

PR #3669, opened 2025-09-19
Handle all exceptions inside intel_openvino

akodanka:
Google's coding guidlines discourage usage of exceptions. This change handles exceptions within the intel_openvino libraries.

### DIFF
--- a/litert/vendors/intel_openvino/dispatch/BUILD
+++ b/litert/vendors/intel_openvino/dispatch/BUILD
@@ -34,7 +34,6 @@ litert_dynamic_lib(
     ],
     copts = [
         "-Os",
-        "-fno-exceptions",
         "-fno-unwind-tables",
         "-fno-asynchronous-unwind-tables",
         "-ffunction-sections",
--- a/litert/vendors/intel_openvino/dispatch/dispatch_api.cc
+++ b/litert/vendors/intel_openvino/dispatch/dispatch_api.cc
@@ -75,13 +75,18 @@ LiteRtStatus DispatchGetCapabilities(int* capabilities) {
 // error.
 LiteRtStatus DispatchDeviceContextCreate(
     LiteRtDispatchDeviceContext* device_context) {
-  if (auto context = LiteRtDispatchDeviceContextT::Create(); context) {
-    *device_context = context->release();
-    return kLiteRtStatusOk;
-  } else {
-    LITERT_LOG(LITERT_ERROR, "Failed to create device context: %s",
-               context.Error().Message().c_str());
-    return context.Error().Status();
+  try {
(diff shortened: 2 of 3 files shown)


litert/vendors/intel_openvino/dispatch/BUILD
```diff
@@ -34,7 +34,6 @@ litert_dynamic_lib(
     ],
     copts = [
         "-Os",
-        "-fno-exceptions",
```
MatthiasKreileder, line 37: Thanks, please add `"-fexceptions",` to the `copts` of the targets under `litert/vendors/intel_openvino/...` so that for the google-internal build exceptions are enabled.

litert/vendors/intel_openvino/dispatch/dispatch_api.cc
```diff
@@ -199,27 +219,37 @@ LiteRtStatus DispatchInvocationContextDestroy(
...
+                                                      tensor_buffer_handle);
+        !status) {
+      LITERT_LOG(LITERT_ERROR, "Failed to attach output: %s",
+                status.Error().Message().c_str());
+      return status.Error().Status();
+    }
+    return kLiteRtStatusOk;
+  } catch (const std::exception& e) {
+    LITERT_LOG(LITERT_ERROR, "Exception in Dispatch attach_output: %s", e.what());
+    return kLiteRtStatusErrorRuntimeFailure;
```
MatthiasKreileder, line 251: Thanks, this is exactly the pattern I had in mind!

I see that you haven't applied this to all function though, e.g. `DispatchDetachInput`. Presumably because you know that the underlying code doesn't throw? Though I am wondering if it would be best to simply wrap "all code", mainly because:

- If the underlying openvino impl changes or someone later changes `DispatchDetachInput` to invoke a function (perhaps indirectly) that might throw, then we'll be in trouble.
- If all functions are ``try/catch`-guarded then the pattern is pretty clear to anyone who makes changes (and perhaps adds new functions).

MatthiasKreileder (commented):
Thanks for the work!

litert/vendors/intel_openvino/dispatch/BUILD
akodanka: Done

litert/vendors/intel_openvino/dispatch/dispatch_api.cc
akodanka: I added exception handling only at required places as it might introduce additional overheads and also clutters the code.

The exception is currently added around all codes that has any calls into the OpenVINO headers/libs.
We'll introduce the exception handlings whenever any new code calls into the Openvino?

PR #3944, opened 2025-10-09
Intel ov options

akodanka:


litert/tools/flags/apply_plugin_flags.cc
```diff
@@ -16,6 +16,7 @@
 
 #include <string>
 
+#define ABSL_FLAGS_STRIP_NAMES 0
```
oToToT: nit: do we need this?

litert/cc/options/litert_intel_openvino_options_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  EXPECT_EQ(options_->GetPerformanceMode(), kLiteRtIntelOpenVinoPerformanceModeCumulativeThroughput);
+}
+
+TEST_F(IntelOpenVinoOptionsTest, SetAndGetConfigsMapOption) {
+  // Set individual config options
+  options_->SetConfigsMapOption("INFERENCE_PRECISION_HINT", "f16");
+  options_->SetConfigsMapOption("NPU_COMPILATION_MODE_PARAMS", "compute-layers-with-higher-precision=Sigmoid");
+  options_->SetConfigsMapOption("CACHE_DIR", "/tmp/ov_cache");
+
+  // Note: GetConfigsMapOption returns a raw pointer to internal C string
```
oToToT: should we call `GetConfigsMapOption` and verify the result here?

litert/tools/flags/vendors/intel_openvino_flags.cc
```diff
@@ -0,0 +1,143 @@
...
+  options.SetPerformanceMode(
+      absl::GetFlag(FLAGS_intel_openvino_performance_mode));
+
+  // Parse configs map options from the flag (comma-separated key=value pairs)
+  const std::string configs_map_str = absl::GetFlag(FLAGS_intel_openvino_configs_map);
+  if (!configs_map_str.empty()) {
+    std::vector<std::string> config_pairs = absl::StrSplit(configs_map_str, ',');
+    for (const auto& pair : config_pairs) {
+      std::vector<std::string> key_value = absl::StrSplit(pair, '=');
+      if (key_value.size() == 2) {
```
oToToT, line 135: nit: maybe we can add some logs if the parsing fail?

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -273,13 +294,84 @@ LiteRtStatus LiteRtCompilerPluginCompile(
...
+          device = "AUTO";
+          break;
+      }
+
+      LITERT_LOG(LITERT_INFO, "Using Intel OpenVINO device: %s", device.c_str());
+
+      // Log other options
+      auto precision = intel_opts.GetInferencePrecision();
+      auto performance_mode = intel_opts.GetPerformanceMode();
+      bool caching_enabled = intel_opts.GetEnableModelCaching();
```
JeevakaPrabu: remove the undefined functions GetInferencePrecision, GetEnableModelCaching and update the log in line no: 328 to reflect this change

litert/tools/flags/apply_plugin_flags.cc
akodanka: This is needed if we need to compile this tool for android.
Are you suggesting to remove this only for apply_plugin since we compile it for the host ?

litert/c/options/litert_intel_openvino_options.cc
```diff
@@ -0,0 +1,164 @@
...
+};
+
+LiteRtStatus LiteRtIntelOpenVinoOptionsCreate(LiteRtOpaqueOptions* options) {
+  if (options == nullptr) {
+    return kLiteRtStatusErrorInvalidArgument;
+  }
+
+  auto options_data = std::make_unique<LiteRtIntelOpenVinoOptionsT>();
+
+  LITERT_RETURN_IF_ERROR(LiteRtCreateOpaqueOptions(
```
MatthiasKreileder, line 45: Please also define and set the function pointer to compute the option hash (`LiteRtSetOpaqueOptionsHash`)  e.g. https://github.com/google-ai-edge/LiteRT/blob/abad9c48ad1dde7c227d25d78e0c6b9ba5ee9d73/litert/c/options/litert_qualcomm_options.cc#L67

Context: This is used for JIT compilation caching, to tell if re-compilation is required.

litert/c/options/litert_intel_openvino_options.h
```diff
@@ -0,0 +1,105 @@
...
+// data can be accessed from the payload.
+LiteRtStatus LiteRtIntelOpenVinoOptionsCreate(LiteRtOpaqueOptions* options);
+LITERT_DEFINE_HANDLE(LiteRtIntelOpenVinoOptions);
+
+// The a string identifier that discriminates Intel OpenVINO options within
+// type erased options.
+const char* LiteRtIntelOpenVinoOptionsGetIdentifier();
+
+// Attempt to retrieve Intel OpenVINO options from the opaque options. Fails
+// unless the opaque options are of another type.
```
fergushenderson: Should that be "if" rather than "unless"?

litert/c/options/litert_intel_openvino_options.h
```diff
@@ -0,0 +1,105 @@
...
+    LiteRtIntelOpenVinoOptions options,
+    LiteRtIntelOpenVinoPerformanceMode performance_mode);
+
+LiteRtStatus LiteRtIntelOpenVinoOptionsGetPerformanceMode(
+    LiteRtIntelOpenVinoOptions options,
+    LiteRtIntelOpenVinoPerformanceMode* performance_mode);
+
+// configs_map ----------------------------------------------------------------
+
+// Set a custom configuration option with a string key-value pair
```
fergushenderson: Consider documenting whether the lifetime of the strings pointed to by `key` and `value` needs to extend beyond the call to this function.

litert/c/options/litert_intel_openvino_options.h
```diff
@@ -0,0 +1,105 @@
...
+LiteRtStatus LiteRtIntelOpenVinoOptionsGetNumConfigsMapOptions(
+    LiteRtIntelOpenVinoOptions options,
+    int* num_options);
+
+// Get a custom configuration option by index
+LiteRtStatus LiteRtIntelOpenVinoOptionsGetConfigsMapOption(
+    LiteRtIntelOpenVinoOptions options,
+    int index,
+    const char** key,
+    const char** value);
```
fergushenderson, line 103: Please document the lifetime of the strings pointed to by `*key` and `*value`.

litert/tools/flags/apply_plugin_flags.cc
oToToT: I'm more on wondering why other vendors do not need this.

MatthiasKreileder: I'm afraid this has to do with the default behaviour of absl flags for mobile, which try to minimize binary size. To prevent the flags from being stripped one can compile like this:

```
bazel build --config=android_x86_64  --copt=-DABSL_FLAGS_STRIP_NAMES=0 --copt=-DABSL_FLAGS_STRIP_HELP=0  //litert/tools:run_model
```

This is arguably not very intuitive and user-friendly. We should probably think to adopt how https://github.com/google-ai-edge/LiteRT/blob/a17305e67ee7874351673f57475858c6dbcc34d1/litert/tools/benchmark_litert_model.h#L348 does this, or maybe even better merge the two tools to avoid confusion.

litert/c/options/litert_intel_openvino_options.h
```diff
@@ -0,0 +1,109 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+// Create an Intel OpenVINO options object that is type erased. The actual option
+// data can be accessed from the payload.
+LiteRtStatus LiteRtIntelOpenVinoOptionsCreate(LiteRtOpaqueOptions* options);
+LITERT_DEFINE_HANDLE(LiteRtIntelOpenVinoOptions);
+
+// The a string identifier that discriminates Intel OpenVINO options within
```
oToToT: The string identifier

litert/c/options/litert_intel_openvino_options.h
```diff
@@ -0,0 +1,109 @@
...
+    const char* value);
+
+// Get the number of custom configuration options
+LiteRtStatus LiteRtIntelOpenVinoOptionsGetNumConfigsMapOptions(
+    LiteRtIntelOpenVinoOptions options,
+    int* num_options);
+
+// Get a custom configuration option by index.
+// The returned key and value pointers are valid for the lifetime of the
+// options object. The caller should not free these pointers.
```
oToToT: If someone Get then Set, the previous options got can be invalidated because the underlying storage is a vector.

litert/tools/flags/apply_plugin_flags.cc
akodanka: I'll remove this commit from the PR, and try using the suggested build flag for resolving the issue.

litert/c/options/litert_intel_openvino_options.h
akodanka: I have changed the cpp implementation to return string copies of the underlying data to avoid this issue.

litert/tools/BUILD
```diff
@@ -110,6 +110,7 @@ litert_bin(
         "//litert/vendors/mediatek/compiler:compiler_plugin_so",
         "//litert/vendors/examples:example_plugin_so",
         "//litert/vendors/qualcomm/compiler:qnn_compiler_plugin_so",
+        "//litert/vendors/intel_openvino/compiler:intel_openvino_compiler_plugin_so",
```
MatthiasKreileder: There are some github action failures, like https://github.com/google-ai-edge/LiteRT/actions/runs/18503064556/job/52783135369?pr=3944#step:9:481 We need to please resolve those to be able to merge.

I had a closer and my understanding of the issue is:

- You are creating a dependency chain from targets like `apply_plugin_main` to `intel_openvino_compiler_plugin_so`. `apply_plugin_main` and other targets in LiteRT get built and tested automatically
- Starting with this pull request bazel needs to build `intel_openvino_compiler_plugin_so`, but fails because there's no way for bazel to download the `@@intel_openvino//:openvino` "external repo".

I can think of two ways how we can continue:

1. Other vendor configure the bazel build like so: https://github.com/google-ai-edge/LiteRT/blob/a6c3c2b81e4cce6b30f39982714eb43b74f01bf7/third_party/neuro_pilot/workspace.bzl#L25 - if you have by now a prebuilt SDK hosted somewhere you could add this to your CL. IIRC there are still some hurdles to clear on your end to host this though?
2. Another solution would be to reduce the scope of this pull request to only add the Intel options, but not hook them into targets like `apply_plugin_main` etc to make the checks pass. Maybe you could stash the additional changes until you have the prebuild SDK so your work isn't lost. If you do this you could maybe also consider splitting this pull request further into the 'fixes' (`Use ov::Tensor for AHWB` + `Use generic ov::Tensor for mapping buffers`) and the commits for the options, but I am also ok leaving this together.

akodanka: I have removed the dependency of intel_openvino_compiler_plugin_so from apply_plugin_main now.
Will reintroduce that once we have the prebuilt sdk.

PR #4268, opened 2025-10-29
Set DMA buffer as the default intel ov delegate

akodanka:
When running gemma3 model using LiteRT-LM, the pipeline shows a huge latency in the cache update model execution.
Since ov::RemoteTensor is currently disabled for AHWB, there was buffer copy involved.
DMA buffer(uses ov::RemoteTensor) helps improve this, but LiteRT throws the error "buffer type is not supported" when the dispatch plugin sets DMA as the default.
This PR adds support for DMA buffer in LiteRT, and sets it as the default from the intel_openvino dispatch delegate.


MatthiasKreileder (approved):
Thanks! Please consider adding a brief PR description to explain the why behind the change.

PR #4435, opened 2025-11-07
Qualcomm AI Engine Direct - Improve init time.

KelvinYang0320:
Summary:
- Introduce maybe_compiled_plugins_ as a member variable to reduce initialization time.

### DIFF
--- a/litert/runtime/compiled_model.h
+++ b/litert/runtime/compiled_model.h
@@ -396,6 +396,7 @@ class LiteRtCompiledModelT {
   // Otherwise, the compiled model will be compiled and saved to the cache.
   std::optional<litert::internal::CompilationCache> compilation_cache_;
   std::optional<LiteRtModelT::Ptr> cached_model_;
+  std::vector<litert::internal::CompilerPlugin> maybe_compiled_plugins_;
 #endif  // !defined(LITERT_DISABLE_NPU)
 
   // The buffer requirement maps for CPU buffers. For delegates with CPU
--- a/litert/runtime/compiled_model.cc
+++ b/litert/runtime/compiled_model.cc
@@ -692,6 +692,9 @@ litert::Expected<bool> LiteRtCompiledModelT::ApplyPluginsWithCaching(
   if (maybe_compiled_plugins.HasValue()) {
     TryApplyPluginsImpl(&model, hw_accelerators, maybe_compiled_plugins.Value(),
                         &need_reserialization);
+    // Store the compiler plugins to a member variable to postpone its
+    // destruction and reduce initialization time.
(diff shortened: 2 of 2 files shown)


weilhuan-qti:
Nit: modify summary to
```
Summary:
- Introduce maybe_compiled_plugins_ as a member variable to postpone its destruction and reduce initialization time.
```

weilhuan-qti (approved):

litert/runtime/compiled_model.h
```diff
@@ -396,6 +390,7 @@ class LiteRtCompiledModelT {
   // Otherwise, the compiled model will be compiled and saved to the cache.
   std::optional<litert::internal::CompilationCache> compilation_cache_;
   std::optional<LiteRtModelT::Ptr> cached_model_;
+  litert::Expected<std::vector<litert::internal::CompilerPlugin>> maybe_compiled_plugins_;
```
YunanAZ: Comments from @terryheo:

litert::Expected<> doesn't look necessary.

YunanAZ (commented):
From @terryheo

As a result of delaying of the destruction, what's the memory consumption overhead introduced?

@KelvinYang0320, could you provide benchmark insights here? Relevant numbers are fine.

litert/runtime/compiled_model.h
```diff
@@ -396,6 +396,7 @@ class LiteRtCompiledModelT {
   // Otherwise, the compiled model will be compiled and saved to the cache.
   std::optional<litert::internal::CompilationCache> compilation_cache_;
   std::optional<LiteRtModelT::Ptr> cached_model_;
+  std::vector<litert::internal::CompilerPlugin> maybe_compiled_plugins_;
```
terryheo, line 399: std::option<> ? Can it be multiple?

KelvinYang0320: > std::option<> ? 

Hi Terry, sorry, but I did not understand the reason for using std::optional in this context. Could you please clarify?

> Can it be multiple?

Typically, there should be only one element in the CompilerPlugin vector. However, if a user uploads more than one file with the "libLiteRtCompilerPlugin" prefix, there could be multiple compiled plugins. I'm not sure if this is intended to prevent unexpected uploads from users.
Could you help clarify the design rationale with the code owner in Google for me?

terryheo: CompiledModel is created from a Model. If users are running two models, two CompiledModels will be used.
In a single CompiledModel, I don't see a case that runs with multiple compilers. If the model is running on a QC devices, only QC's CompilerPlugin is needed. So I don't think vector<> is not needed.

PR #4437, opened 2025-11-07
Add more ops to intel_openvino supported list

akodanka:


third_party/intel_openvino/openvino.bazel
```diff
@@ -23,7 +23,6 @@ cc_library(
         "//conditions:default": [
             "openvino/runtime/lib/intel64/libopenvino.so",
             "openvino/runtime/lib/intel64/libopenvino_tensorflow_lite_frontend.so",
-            "openvino/runtime/lib/intel64/libc++_shared.so",
```
MatthiasKreileder, line 26: Just for my understanding: Why has this been removed?

It was added in https://github.com/google-ai-edge/LiteRT/pull/2620/files#diff-077ca974b323610126337e52823befa5780f41d50d4f365d5df42b6f19d0dd11

PR #4466, opened 2025-11-10
Qualcomm AI Engine Direct - Parse and add vendor options.

KelvinYang0320:
Summary:
- Fix missing call to ParseAndAddOptions for vendor-specific options in apply_plugin_main.

### DIFF
--- a/litert/tools/apply_plugin_main.cc
+++ b/litert/tools/apply_plugin_main.cc
@@ -144,13 +144,13 @@ int main(int argc, char* argv[]) {
 
   bool all_flags_parsed =
       (ParseAndAddOptions<litert::CompilerOptions>(
-           *opts, run, litert::CompilerOptionsFromFlags) ||
+           *opts, run, litert::CompilerOptionsFromFlags) &&
        ParseAndAddOptions<litert::qualcomm::QualcommOptions>(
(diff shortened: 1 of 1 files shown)


litert/tools/apply_plugin_main.cc
```diff
@@ -142,21 +142,24 @@ int main(int argc, char* argv[]) {
...
   }
+  bool vendor_flags_parsed =
+      ParseAndAddOptions<litert::qualcomm::QualcommOptions>(
+          *opts, run, litert::qualcomm::QualcommOptionsFromFlags) ||
+      ParseAndAddOptions<litert::google_tensor::GoogleTensorOptions>(
+          *opts, run, litert::google_tensor::GoogleTensorOptionsFromFlags) ||
+      ParseAndAddOptions<litert::intel_openvino::IntelOpenVinoOptions>(
+          *opts, run, litert::intel_openvino::IntelOpenVinoOptionsFromFlags) ||
+      ParseAndAddOptions<litert::mediatek::MediatekOptions>(
+          *opts, run, litert::mediatek::MediatekOptionsFromFlags);
```
weilhuan-qti: I think the correct fix is to call every parse function and "and" their result (instead of "or" their result).

weilhuan-qti (approved):

PR #4737, opened 2025-11-24
Updated build_ios.md

aravindmurali22:
changed bazel build to correct ios target paths in build_ios.md: `//tensorflow/lite/ `to `//tflite/`


junjiang-lab:
Thanks Aravind for the update.

But we need to update copybara config to transform those paths since this `build_ios.md` file is under tensorflow/lite codes.

PR #4753, opened 2025-11-26
Add DirectX 12 tensor buffer to share memory with OpenVINO NPU on Windows

fujunwei:


fujunwei:
Thanks for taking a high-level look at the design and supporting this buffer type, i’ll refine the code and break it down into smaller pull requests to streamline your review process.

PR #4754, opened 2025-11-26
Qualcomm AI Engine Direct - Preserve converted data.

KelvinYang0320:
Summary:
- Fix invalid data issue after int4-to-int8 conversion.
- Add unit tests for TensorWrapper construction to ensure correct data handling.

### DIFF
--- a/litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
+++ b/litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
@@ -12,6 +12,7 @@
 #include <variant>
 #include <vector>
 
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include "absl/types/span.h"  // from @com_google_absl
 #include "litert/vendors/qualcomm/core/utils/miscs.h"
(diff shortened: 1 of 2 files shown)


litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -457,61 +458,123 @@ TEST(TensorWrapperTest, ConstQnnTensorPerTensorQuantConstructTest) {
...
+  const auto* tensor_data_ptr = static_cast<const T*>(tensor.v2.clientBuf.data);
+  EXPECT_THAT(
+      std::vector<T>(tensor_data_ptr, tensor_data_ptr + expected_data.size()),
+      ::testing::ElementsAreArray(expected_data));
+}
+
+template <typename T, bool is_int4 = false>
+void RunQnnTensorImplicitCopyTest(Qnn_DataType_t datatype) {
+  std::vector<std::uint32_t> dims = {1, 1, 4};
+  std::vector<T> data = {1, 2, 3, 4};
```
chunhsue-qti: Should we put these constants in anonymous namespace and rename them into `kDims` and `kData`?

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -457,61 +458,123 @@ TEST(TensorWrapperTest, ConstQnnTensorPerTensorQuantConstructTest) {
...
+  if constexpr (is_int4) {
+    std::vector<int8_t> packed_data;
+    packed_data.reserve(data.size() / 2);
+    for (size_t i = 0; i < data.size() / 2; ++i) {
+      uint8_t low_nibble = data[i * 2] & 0x0F;
+      uint8_t high_nibble = data[i * 2 + 1] & 0x0F;
+      packed_data.emplace_back((high_nibble << 4) | low_nibble);
```
chunhsue-qti, line 499: NIT: 
Make this a util functions so it can be reused in the future.

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
KelvinYang0320: Thank you. I have updated `dims` to `kDims`.
As for `data`, the function will modify its value depending on `is_int4`, so I will leave it unchanged.

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
KelvinYang0320: Thank you for the suggestion, but I would like to create the utility function after we have a clear requirement.
I prefer to keep this PR small.

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -457,61 +458,124 @@ TEST(TensorWrapperTest, ConstQnnTensorPerTensorQuantConstructTest) {
...
-  ASSERT_EQ(ref.v2.clientBuf.dataSize, data_size);
-  const auto* ref_data =
-      static_cast<const std::uint8_t*>(ref.v2.clientBuf.data);
-  for (size_t i = 0; i < data.size(); i++) {
-    EXPECT_EQ(ref_data[i], data[i]);
+}
+
+TEST_P(TensorWrapperDatatypeTest, QnnTensorImplicitCopyTest) {
+  const Qnn_DataType_t datatype = GetParam();
+  switch (datatype) {
```
weilhuan-qti: In our current use cases, I think we don't need parameterized tests here because we check the parameter and do different things.

Directly run `RunQnnTensorImplicitCopyTest` with type we want to test.
```suggestion
    RunQnnTensorImplicitCopyTest<std::uint8_t>(QNN_DATATYPE_UFIXED_POINT_8);
    RunQnnTensorImplicitCopyTest<std::int8_t>(QNN_DATATYPE_SFIXED_POINT_8);
    RunQnnTensorImplicitCopyTest<std::uint16_t>(QNN_DATATYPE_UFIXED_POINT_16);
    RunQnnTensorImplicitCopyTest<std::int16_t>(QNN_DATATYPE_SFIXED_POINT_16);
    RunQnnTensorImplicitCopyTest<std::int8_t, true>(QNN_DATATYPE_SFIXED_POINT_4);
```

KelvinYang0320: I would prefer to keep the parameterized tests, as this allows us to quickly identify datatype failures through the task name and easily reproduce them using the gtest filter.
For instance, if someone uses QNN_DATATYPE_SFIXED_POINT_4 incorrectly again, you will be able to immediately identify that there is an issue with this datatype (SFIXED_POINT_4).

KelvinYang0320: After offline discussion, I have changed to 5 `TEST(TensorWrapperDatatypeTest, <datatype>)` and removed `TEST_P`.
Thank you!

weilhuan-qti (approved):

PR #4756, opened 2025-11-26
Qualcomm AI Engine Direct - Add comparison functions for wrappers.

graham0824:
Summary:
- Add comparison operator for wrappers.
- Enable related test for all wrappers.

Test Result:
======================== Test Summary ========================
//litert/c:litert_op_options_test

//litert/c/options:litert_qualcomm_options_test
//litert/c/options:litert_qualcomm_options_test                          PASSED in 0.0s

//litert/tools/flags/vendors:qualcomm_flags_test
//litert/tools/flags/vendors:qualcomm_flags_test                         PASSED in 0.0s

//litert/vendors/qualcomm/core/utils:utils_test
//litert/vendors/qualcomm/core/utils:utils_test                          PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:op_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:op_wrapper_test            PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:tensor_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:tensor_wrapper_test        PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:param_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:param_wrapper_test         PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:quantize_params_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:quantize_params_wrapper_test PASSED in 0.0s

//litert/vendors/qualcomm/core:common_test
//litert/vendors/qualcomm/core:common_test                               PASSED in 0.0s

//litert/vendors/qualcomm/core:tensor_pool_test
//litert/vendors/qualcomm/core:tensor_pool_test                          PASSED in 0.0s

//litert/vendors/qualcomm/core/transformation:graph_to_graph_test
//litert/vendors/qualcomm/core/transformation:graph_to_graph_test        PASSED in 0.0s

//litert/vendors/qualcomm/qnn_backend_test:all
//litert/vendors/qualcomm/qnn_backend_test:qnn_model_test                PASSED in 0.5s

//litert/vendors/qualcomm/qnn_backend_test/builder_test:all
//litert/vendors/qualcomm/qnn_backend_test/builder_test:fully_connected_int2_test PASSED in 0.5s
//litert/vendors/qualcomm/qnn_backend_test/builder_test:relu_test        PASSED in 0.5s
//litert/vendors/qualcomm/qnn_backend_test/builder_test:topk_test        PASSED in 0.5s

//litert/vendors/qualcomm/core/dump:dump_graph_test
//litert/vendors/qualcomm/core/dump:dump_graph_test                      PASSED in 0.0s

//litert/vendors/qualcomm:qnn_manager_test
//litert/vendors/qualcomm:qnn_manager_test                               PASSED in 0.3s

//litert/vendors/qualcomm/compiler:qnn_compiler_plugin_test
//litert/vendors/qualcomm/compiler:qnn_compiler_plugin_test              PASSED in 56.2s

### DIFF
--- a/litert/vendors/qualcomm/core/utils/miscs.h
+++ b/litert/vendors/qualcomm/core/utils/miscs.h
@@ -10,6 +10,7 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <filesystem>
 #include <limits>
 #include <memory>
@@ -23,6 +24,17 @@
 #include "QnnInterface.h"  // from @qairt
 #include "QnnTypes.h"  // from @qairt
 namespace qnn {
+namespace {
+
+inline bool IsStrEq(const char* input, const char* golden) {
+  if (input == nullptr && golden == nullptr) return true;
+  if (input == nullptr || golden == nullptr) return false;
+
+  return std::strcmp(input, golden) == 0;
+}
+
+}  // namespace
+
 constexpr uint32_t kUint16ZeroPoint = -std::numeric_limits<std::int16_t>::min();
 constexpr uint32_t kQuantBitWidth4 = 4;
 constexpr uint32_t kQuantBitWidth2 = 2;
--- a/litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
+++ b/litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
@@ -3,6 +3,7 @@
 
 #include "litert/vendors/qualcomm/core/wrappers/op_wrapper.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <functional>
 #include <optional>
@@ -32,6 +33,28 @@ OpWrapper& OpWrapper::operator=(const OpWrapper& other) {
     new (this) OpWrapper(other);
   }
   return *this;
+}
+
+bool OpWrapper::operator==(const OpWrapper& other) const {
+  if (op_code_ != other.op_code_) return false;
+  if (!IsStrEq(type_name_, other.type_name_)) return false;
+
+  if (!std::equal(
+          input_tensors_.begin(), input_tensors_.end(),
+          other.input_tensors_.begin(), other.input_tensors_.end(),
+          [](const auto& a, const auto& b) { return a.get() == b.get(); })) {
+    return false;
+  }
+
+  if (!std::equal(
+          output_tensors_.begin(), output_tensors_.end(),
+          other.output_tensors_.begin(), other.output_tensors_.end(),
+          [](const auto& a, const auto& b) { return a.get() == b.get(); })) {
+    return false;
+  }
+
+  return scalar_params_ == other.scalar_params_ &&
+         tensor_params_ == other.tensor_params_;
 }
 
 OpWrapper::OpWrapper(OpWrapper&& other)
--- a/litert/vendors/qualcomm/core/wrappers/op_wrapper.h
+++ b/litert/vendors/qualcomm/core/wrappers/op_wrapper.h
@@ -26,6 +26,8 @@ class OpWrapper final {
 
   OpWrapper& operator=(const OpWrapper& other);
 
+  bool operator==(const OpWrapper& other) const;
+
   OpWrapper(OpWrapper&& other);
 
   ~OpWrapper();
--- a/litert/vendors/qualcomm/core/wrappers/param_wrapper.h
+++ b/litert/vendors/qualcomm/core/wrappers/param_wrapper.h
@@ -55,6 +55,8 @@ class ScalarParamWrapper {
     }
   }
 
+  bool operator==(const ScalarParamWrapper& other) const;
+
   void CloneTo(Qnn_Param_t& dst) const;
 
  private:
@@ -65,6 +67,8 @@ class ScalarParamWrapper {
 class TensorParamWrapper {
  public:
   explicit TensorParamWrapper(const char* name, const TensorWrapper& tensor);
+
+  bool operator==(const TensorParamWrapper& other) const;
 
   void CloneTo(Qnn_Param_t& dst) const;
 
--- a/litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
+++ b/litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
@@ -2,6 +2,7 @@
 // All Rights Reserved.
 #include "litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
@@ -20,6 +21,14 @@ UndefinedQuantizeParamsWrapper::UndefinedQuantizeParamsWrapper(
 
 UndefinedQuantizeParamsWrapper::UndefinedQuantizeParamsWrapper(
     UndefinedQuantizeParamsWrapper&&) = default;
+
+bool UndefinedQuantizeParamsWrapper::operator==(
+    const UndefinedQuantizeParamsWrapper& other) const {
+  return qnn_quantize_param_.encodingDefinition ==
+             other.qnn_quantize_param_.encodingDefinition &&
+         qnn_quantize_param_.quantizationEncoding ==
+             other.qnn_quantize_param_.quantizationEncoding;
+}
 
 void UndefinedQuantizeParamsWrapper::CloneTo(Qnn_QuantizeParams_t& dst) {
   dst = qnn_quantize_param_;
@@ -48,6 +57,17 @@ ScaleOffsetQuantizeParamsWrapper::ScaleOffsetQuantizeParamsWrapper(
 ScaleOffsetQuantizeParamsWrapper::ScaleOffsetQuantizeParamsWrapper(
     ScaleOffsetQuantizeParamsWrapper&&) = default;
 
+bool ScaleOffsetQuantizeParamsWrapper::operator==(
+    const ScaleOffsetQuantizeParamsWrapper& other) const {
+  return qnn_quantize_param_.encodingDefinition ==
+             other.qnn_quantize_param_.encodingDefinition &&
+         qnn_quantize_param_.quantizationEncoding ==
+             other.qnn_quantize_param_.quantizationEncoding &&
+         qnn_quantize_param_.scaleOffsetEncoding.scale ==
+             other.qnn_quantize_param_.scaleOffsetEncoding.scale &&
+         qnn_quantize_param_.scaleOffsetEncoding.offset ==
+             other.qnn_quantize_param_.scaleOffsetEncoding.offset;
+}
 void ScaleOffsetQuantizeParamsWrapper::CloneTo(Qnn_QuantizeParams_t& dst) {
   dst = qnn_quantize_param_;
 }
@@ -105,6 +125,30 @@ AxisScaleOffsetQuantizeParamsWrapper::AxisScaleOffsetQuantizeParamsWrapper(
       scale_offsets_.data();
 }
 
+bool AxisScaleOffsetQuantizeParamsWrapper::operator==(
+    const AxisScaleOffsetQuantizeParamsWrapper& other) const {
+  if (qnn_quantize_param_.encodingDefinition !=
+      other.qnn_quantize_param_.encodingDefinition)
+    return false;
+  if (qnn_quantize_param_.quantizationEncoding !=
+      other.qnn_quantize_param_.quantizationEncoding)
+    return false;
+  if (qnn_quantize_param_.axisScaleOffsetEncoding.axis !=
+      other.qnn_quantize_param_.axisScaleOffsetEncoding.axis)
+    return false;
+  if (qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets !=
+      other.qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets)
+    return false;
+
+  return std::equal(
+      qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+      qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset +
+          qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
+      other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+      [](const Qnn_ScaleOffset_t& a, const Qnn_ScaleOffset_t& b) {
+        return a.scale == b.scale && a.offset == b.offset;
+      });
+}
 void AxisScaleOffsetQuantizeParamsWrapper::CloneTo(Qnn_QuantizeParams_t& dst) {
   dst = qnn_quantize_param_;
 }
@@ -152,6 +196,20 @@ BwScaleOffsetQuantizeParamsWrapper::BwScaleOffsetQuantizeParamsWrapper(
 BwScaleOffsetQuantizeParamsWrapper::BwScaleOffsetQuantizeParamsWrapper(
     BwScaleOffsetQuantizeParamsWrapper&& rhs) = default;
 
+bool BwScaleOffsetQuantizeParamsWrapper::operator==(
+    const BwScaleOffsetQuantizeParamsWrapper& other) const {
+  return qnn_quantize_param_.encodingDefinition ==
+             other.qnn_quantize_param_.encodingDefinition &&
+         qnn_quantize_param_.quantizationEncoding ==
+             other.qnn_quantize_param_.quantizationEncoding &&
+         qnn_quantize_param_.bwScaleOffsetEncoding.bitwidth ==
+             other.qnn_quantize_param_.bwScaleOffsetEncoding.bitwidth &&
+         qnn_quantize_param_.bwScaleOffsetEncoding.scale ==
+             other.qnn_quantize_param_.bwScaleOffsetEncoding.scale &&
+         qnn_quantize_param_.bwScaleOffsetEncoding.offset ==
+             other.qnn_quantize_param_.bwScaleOffsetEncoding.offset;
+}
+
 void BwScaleOffsetQuantizeParamsWrapper::CloneTo(Qnn_QuantizeParams_t& dst) {
   dst = qnn_quantize_param_;
 }
@@ -192,6 +250,36 @@ BwAxisScaleOffsetQuantizeParamsWrapper::BwAxisScaleOffsetQuantizeParamsWrapper(
       offsets_{std::move(rhs.offsets_)} {
   qnn_quantize_param_.bwAxisScaleOffsetEncoding.scales = scales_.data();
   qnn_quantize_param_.bwAxisScaleOffsetEncoding.offsets = offsets_.data();
+}
+
+bool BwAxisScaleOffsetQuantizeParamsWrapper::operator==(
+    const BwAxisScaleOffsetQuantizeParamsWrapper& other) const {
+  if (qnn_quantize_param_.encodingDefinition !=
+      other.qnn_quantize_param_.encodingDefinition)
+    return false;
+  if (qnn_quantize_param_.quantizationEncoding !=
+      other.qnn_quantize_param_.quantizationEncoding)
+    return false;
+  if (qnn_quantize_param_.bwAxisScaleOffsetEncoding.bitwidth !=
+      other.qnn_quantize_param_.bwAxisScaleOffsetEncoding.bitwidth)
+    return false;
+  if (qnn_quantize_param_.bwAxisScaleOffsetEncoding.axis !=
+      other.qnn_quantize_param_.bwAxisScaleOffsetEncoding.axis)
+    return false;
+  if (qnn_quantize_param_.bwAxisScaleOffsetEncoding.numElements !=
+      other.qnn_quantize_param_.bwAxisScaleOffsetEncoding.numElements)
+    return false;
+
+  return std::equal(
+             qnn_quantize_param_.bwAxisScaleOffsetEncoding.scales,
+             qnn_quantize_param_.bwAxisScaleOffsetEncoding.scales +
+                 qnn_quantize_param_.bwAxisScaleOffsetEncoding.numElements,
+             other.qnn_quantize_param_.bwAxisScaleOffsetEncoding.scales) &&
+         std::equal(
+             qnn_quantize_param_.bwAxisScaleOffsetEncoding.offsets,
+             qnn_quantize_param_.bwAxisScaleOffsetEncoding.offsets +
+                 qnn_quantize_param_.bwAxisScaleOffsetEncoding.numElements,
+             other.qnn_quantize_param_.bwAxisScaleOffsetEncoding.offsets);
 }
 
 void BwAxisScaleOffsetQuantizeParamsWrapper::CloneTo(
--- a/litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h
+++ b/litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h
@@ -21,6 +21,8 @@ class UndefinedQuantizeParamsWrapper final {
 
   UndefinedQuantizeParamsWrapper(UndefinedQuantizeParamsWrapper&&);
 
+  bool operator==(const UndefinedQuantizeParamsWrapper& other) const;
+
   void CloneTo(Qnn_QuantizeParams_t& dst);
 
  private:
@@ -37,6 +39,8 @@ class ScaleOffsetQuantizeParamsWrapper final {
   ScaleOffsetQuantizeParamsWrapper(const ScaleOffsetQuantizeParamsWrapper&);
 
   ScaleOffsetQuantizeParamsWrapper(ScaleOffsetQuantizeParamsWrapper&&);
+
+  bool operator==(const ScaleOffsetQuantizeParamsWrapper& other) const;
 
   void CloneTo(Qnn_QuantizeParams_t& dst);
 
@@ -71,6 +75,8 @@ class AxisScaleOffsetQuantizeParamsWrapper final {
   AxisScaleOffsetQuantizeParamsWrapper(
       AxisScaleOffsetQuantizeParamsWrapper&& rhs);
 
+  bool operator==(const AxisScaleOffsetQuantizeParamsWrapper& other) const;
+
   void CloneTo(Qnn_QuantizeParams_t& dst);
 
   std::int32_t GetAxis() const;
@@ -97,6 +103,8 @@ class BwScaleOffsetQuantizeParamsWrapper final {
 
   BwScaleOffsetQuantizeParamsWrapper(BwScaleOffsetQuantizeParamsWrapper&& rhs);
 
+  bool operator==(const BwScaleOffsetQuantizeParamsWrapper& other) const;
+
   void CloneTo(Qnn_QuantizeParams_t& dst);
 
   std::uint32_t GetBitwidth() const {
@@ -119,6 +127,8 @@ class BwAxisScaleOffsetQuantizeParamsWrapper final {
 
   BwAxisScaleOffsetQuantizeParamsWrapper(
       BwAxisScaleOffsetQuantizeParamsWrapper&& rhs);
+
+  bool operator==(const BwAxisScaleOffsetQuantizeParamsWrapper& other) const;
 
   void CloneTo(Qnn_QuantizeParams_t& dst);
 
--- a/litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
+++ b/litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
@@ -112,9 +112,11 @@ class TensorWrapper final {
 
   ~TensorWrapper();
 
-  bool operator==(const TensorWrapper& other) const { return this == &other; }
+  bool operator==(const TensorWrapper& other) const;
 
-  bool operator!=(const TensorWrapper& other) const { return this != &other; }
+  bool operator!=(const TensorWrapper& other) const {
+    return !(*this == other);
+  }
 
   void CloneTo(Qnn_Tensor_t& dst) const;
 
--- a/litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
+++ b/litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
@@ -209,5 +209,219 @@ TEST(OpWrapperTest, ChangeOpName) {
   EXPECT_STREQ(op_wrapper_1.GetOpConfig().v1.name, "namespace/name_new");
 }
 
+TensorWrapper CreateTensor(const std::vector<uint32_t>& dims,
+                           Qnn_TensorType_t type,
+                           std::optional<uint8_t> val = std::nullopt) {
+  std::vector<uint8_t> data(
+      std::accumulate(dims.begin(), dims.end(), 1u, std::multiplies<>()));
+  if (val.has_value()) {
+    std::fill(data.begin(), data.end(), *val);
+  }
+  return TensorWrapper("", type, QNN_DATATYPE_UFIXED_POINT_8,
+                       QuantizeParamsWrapperVariant(), dims,
+                       static_cast<uint32_t>(data.size()), data.data(), true);
+}
+
+TEST(OpWrapperEqualityOperatorTest, TypeName) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, QNN_TENSOR_TYPE_NATIVE);
+  auto output = input;
+  auto param_tensor = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 1);
+
+  qnn::OpWrapper op1{"op", "TYPE_A", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "TYPE_A", QnnOpCode::kUnknown};
+  op2.AddInputTensor(input);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "TYPE_B", QnnOpCode::kUnknown};
+  op3.AddInputTensor(input);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 255, false);
+  op3.AddTensorParam("tensor_param", param_tensor);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
+TEST(OpWrapperEqualityOperatorTest, OpCode) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, QNN_TENSOR_TYPE_NATIVE);
+  auto output = input;
+  auto param_tensor = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 1);
+
+  qnn::OpWrapper op1{"op", "OP_TYPE", QnnOpCode::kElementWiseAdd};
+  op1.AddInputTensor(input);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "OP_TYPE", QnnOpCode::kElementWiseAdd};
+  op2.AddInputTensor(input);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "OP_TYPE", QnnOpCode::kElementWiseMultiply};
+  op3.AddInputTensor(input);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 255, false);
+  op3.AddTensorParam("tensor_param", param_tensor);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
+TEST(OpWrapperEqualityOperatorTest, InputDims) {
+  std::vector<uint32_t> dims1 = {1, 1, 3};
+  std::vector<uint32_t> dims2 = {1, 1, 4};
+  auto input1 = CreateTensor(dims1, QNN_TENSOR_TYPE_NATIVE);
+  auto input2 = CreateTensor(dims2, QNN_TENSOR_TYPE_NATIVE);
+  auto output = input1;
+  auto param_tensor = CreateTensor(dims1, QNN_TENSOR_TYPE_STATIC, 1);
+
+  qnn::OpWrapper op1{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input1);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op2.AddInputTensor(input1);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op3.AddInputTensor(input2);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 255, false);
+  op3.AddTensorParam("tensor_param", param_tensor);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
+TEST(OpWrapperEqualityOperatorTest, ScalarParam) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, QNN_TENSOR_TYPE_NATIVE);
+  auto output = input;
+  auto param_tensor = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 1);
+
+  qnn::OpWrapper op1{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op2.AddInputTensor(input);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op3.AddInputTensor(input);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 252, false);
+  op3.AddTensorParam("tensor_param", param_tensor);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
+TEST(OpWrapperEqualityOperatorTest, TensorParam) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, QNN_TENSOR_TYPE_NATIVE);
+  auto output = input;
+  auto param_tensor = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 1);
+  auto param_tensor_diff = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 2);
+
+  qnn::OpWrapper op1{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op2.AddInputTensor(input);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op3.AddInputTensor(input);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 255, false);
+  op3.AddTensorParam("tensor_param", param_tensor_diff);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
+TEST(OpWrapperEqualityOperatorTest, InputSize) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, QNN_TENSOR_TYPE_NATIVE);
+  auto output = input;
+  auto param_tensor = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 1);
+
+  qnn::OpWrapper op1{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op2.AddInputTensor(input);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op3.AddInputTensor(input);
+  op3.AddInputTensor(input);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 255, false);
+  op3.AddTensorParam("tensor_param", param_tensor);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
+TEST(OpWrapperEqualityOperatorTest, InputData) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input1 = CreateTensor(dims, QNN_TENSOR_TYPE_APP_WRITE, 1);
+  auto input2 = CreateTensor(dims, QNN_TENSOR_TYPE_APP_WRITE, 2);
+  auto output = CreateTensor(dims, QNN_TENSOR_TYPE_APP_READ, 2);
+  auto param_tensor = CreateTensor(dims, QNN_TENSOR_TYPE_STATIC, 1);
+
+  qnn::OpWrapper op1{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input1);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam<std::uint8_t>("param", 255, false);
+  op1.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op2{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op2.AddInputTensor(input1);
+  op2.AddOutputTensor(output);
+  op2.AddScalarParam<std::uint8_t>("param", 255, false);
+  op2.AddTensorParam("tensor_param", param_tensor);
+
+  qnn::OpWrapper op3{"op", "OP_TYPE", QnnOpCode::kUnknown};
+  op3.AddInputTensor(input2);
+  op3.AddOutputTensor(output);
+  op3.AddScalarParam<std::uint8_t>("param", 255, false);
+  op3.AddTensorParam("tensor_param", param_tensor);
+
+  EXPECT_TRUE(op1 == op2);
+  EXPECT_FALSE(op1 == op3);
+}
+
 }  // namespace
 }  // namespace qnn
--- a/litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
+++ b/litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
@@ -186,6 +186,55 @@ TEST(ScalarParamWrapperTest, QuantizedInt32ParamTest) {
   EXPECT_EQ(int32_quant_qnn_param.scalarParam.int32Value, value);
 }
 
+template <typename T, bool IsQuant>
+struct ScalarParamType {
+  using Type = T;
+  static constexpr bool is_quant = IsQuant;
+};
+
+template <typename T>
+class ScalarParamEqualTypedTest : public ::testing::Test {};
+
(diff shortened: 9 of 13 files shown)


litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -112,7 +112,9 @@ class TensorWrapper final {
 
   ~TensorWrapper();
 
-  bool operator==(const TensorWrapper& other) const { return this == &other; }
+  bool operator==(const TensorWrapper& other) const {
+    return CompareTensor(qnn_tensor_, other.qnn_tensor_);
```
weilhuan-qti: We can still check the address first.
```suggestion
    return this == &other || CompareTensor(qnn_tensor_, other.qnn_tensor_);
```

litert/vendors/qualcomm/core/wrappers/param_wrapper.h
```diff
@@ -55,6 +55,10 @@ class ScalarParamWrapper {
     }
   }
 
+  bool operator==(const ScalarParamWrapper& other) const {
+    return CompareScalarParam(qnn_scalar_, other.qnn_scalar_);
```
weilhuan-qti: If the design of `*Wrapper` is to hide the QNN struct, we should only manipulate the QNN struct inside `*Wrapper`. For others, just use `*Wrapper::operator==`.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -77,5 +112,150 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
+ * This is useful for testing and debugging.
+ * The comparison is done by calling the Compare* functions above.
+ */
+bool operator==(const Qnn_OpConfigV1_t&, const Qnn_OpConfigV1_t&);
+
+bool operator==(const Qnn_Param_t&, const Qnn_Param_t&);
+
+bool operator==(const Qnn_Scalar_t&, const Qnn_Scalar_t&);
+
+bool operator==(const Qnn_Tensor_t&, const Qnn_Tensor_t&);
```
weilhuan-qti: I am not sure whether it is a good practice to have such `operator==` in the `qnn` namespace. But I think the parameters should be `*Wrapper` instead of QNN struct.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -77,5 +112,150 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
+  CHECK_VALUE_EQ(input.v2.sparseParams.hybridCoo.numSparseDimensions,
+                 golden.v2.sparseParams.hybridCoo.numSparseDimensions,
+                 "Number of sparse dimensions from sparse params");
+  CHECK_TYPE_EQ(input.v2.sparseParams.type, golden.v2.sparseParams.type,
+                "Type of sparse params");
+  return true;
+}
+
+inline bool CompareScalarParam(const Qnn_Scalar_t& input,
+                               const Qnn_Scalar_t& golden) {
```
weilhuan-qti: If we want to hide QNN struct, this implementation should be moved inside `ScalarParamWrapper::operator==`.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -20,9 +20,44 @@
...
 #include "QnnTypes.h"  // from @qairt
 namespace qnn {
+namespace {
+#define CHECK_STR_EQ(INPUT, GOLDEN, MSG)                                       \
+  do {                                                                         \
+    const char* _input = (INPUT);                                              \
+    const char* _golden = (GOLDEN);                                            \
+    if ((_input == nullptr && _golden != nullptr) ||                           \
+        (_input != nullptr && _golden == nullptr) ||                           \
+        (_input && _golden && std::strcmp(_input, _golden) != 0)) {            \
```
weilhuan-qti: Maybe `std::strcmp` already handle nullptr then we don't need to manually check.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -77,5 +112,150 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
+  CHECK_VALUE_EQ(input.numOfOutputs, golden.numOfOutputs,
+                 "Number of op outputs");
+  return true;
+}
+
+inline bool CompareTensor(const Qnn_Tensor_t& input,
+                          const Qnn_Tensor_t& golden) {
+  CHECK_VALUE_EQ(input.version, golden.version, "Tensor version");
+  CHECK_VALUE_EQ(input.v2.id, golden.v2.id, "Id version");
+  CHECK_STR_EQ(input.v2.name, golden.v2.name, "Tensor name");
```
weilhuan-qti: I think we don't need to check id and name because they should be unique across tensors.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -77,5 +112,150 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
+  CHECK_VALUE_EQ(input.v2.isProduced, golden.v2.isProduced,
+                 "Tensor.isProduced");
+
+  // Quantize Param
+  CHECK_VALUE_EQ(input.v2.quantizeParams.encodingDefinition,
+                 golden.v2.quantizeParams.encodingDefinition,
+                 "Encoding definition of quantize params");
+  CHECK_VALUE_EQ(input.v2.quantizeParams.quantizationEncoding,
+                 golden.v2.quantizeParams.quantizationEncoding,
+                 "Quantization encoding of quantize params");
```
weilhuan-qti: Should handle this in `QuantizeParamsWrapper`.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -77,5 +112,150 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
+  CHECK_TYPE_EQ(input.v2.sparseParams.type, golden.v2.sparseParams.type,
+                "Type of sparse params");
+  CHECK_VALUE_EQ(input.v2.sparseParams.hybridCoo.numSpecifiedElements,
+                 golden.v2.sparseParams.hybridCoo.numSpecifiedElements,
+                 "Number of specified elements from sparse params");
+  CHECK_VALUE_EQ(input.v2.sparseParams.hybridCoo.numSparseDimensions,
+                 golden.v2.sparseParams.hybridCoo.numSparseDimensions,
+                 "Number of sparse dimensions from sparse params");
+  CHECK_TYPE_EQ(input.v2.sparseParams.type, golden.v2.sparseParams.type,
+                "Type of sparse params");
```
weilhuan-qti: Maybe currently we don't need to check this and dynamic dimension because we don't use them.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -77,5 +112,150 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
+                          const Qnn_OpConfigV1_t& golden) {
+  CHECK_STR_EQ(input.name, golden.name, "Op name");
+  CHECK_STR_EQ(input.packageName, golden.packageName, "Op package name");
+  CHECK_STR_EQ(input.typeName, golden.typeName, "Op type name");
+  CHECK_VALUE_EQ(input.numOfParams, golden.numOfParams, "Number of op params");
+  CHECK_VALUE_EQ(input.numOfInputs, golden.numOfInputs, "Number of op inputs");
+  CHECK_VALUE_EQ(input.numOfOutputs, golden.numOfOutputs,
+                 "Number of op outputs");
+  return true;
+}
```
weilhuan-qti: Move implementation into `OpWrapper::operator==`. And I think we don't need to check `name`.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
graham0824: Thanks for the comment! Updated in the latest commit.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: After checking the source code of [strcmp](https://github.com/lattera/glibc/blob/master/string/strcmp.c#L30), it will not check if the input is nullptr, and entering nullptr will trigger undefined behavior. So I think we should keep this code.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: You're right! Thanks for the reminder, let me remove these comparisons.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: In the current macro design, if this feature is not used or is set in both the input and golden, the macro will return true, so those comparisons won't be triggered even if we keep the code.
However, your suggestion makes sense. We can add them later when we have an actual use case for comparison. I'll remove them for now, thanks!

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Agree, it's better to hide QNN struct to only compare wrapper, updated in the latest commit.

litert/vendors/qualcomm/core/wrappers/param_wrapper.h
graham0824: Thanks for the comment! Refactor to only do the comparison inside the *Wrapper::operator==, updated in the latest commit.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Thanks for the comment! Although it belongs to qnn namespace, I realized that exposing the way we compare the QNN struct violates the design principle of the *Wrapper. Removed all of them and updated in the latest commit.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Thanks for the reminder! Move into QuantizeParamsWrapper and extend to add "operator==" for all *QuantizeParamsWrapper, updated in the latest commit.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: In the current OpWrapper design, the OpConfig will be populated after calling GetOpConfig(), I realized that we don't need to compare the OpConfig inside the OpWrapper. Maybe we could keep this here for test usage, let's discuss it offline.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -20,9 +20,44 @@
...
+                    _input ? _input : "(null)", _golden ? _golden : "(null)"); \
+      return false;                                                            \
+    }                                                                          \
+  } while (0)
+
+#define CHECK_VALUE_EQ(INPUT, GOLDEN, MSG)                         \
+  do {                                                             \
+    if (INPUT != GOLDEN) {                                         \
+      QNN_LOG_ERROR("%s mismatch. Input: %lld, Golden: %lld", MSG, \
+                    (long long)(INPUT), (long long)(GOLDEN));      \
```
weilhuan-qti: Prefer using `static_cast` instead of c-style cast.

weilhuan-qti: I am thinking that should we print log for every check. I know it help debugging but it also make `operator==` complicate.

litert/vendors/qualcomm/core/wrappers/param_wrapper.h
```diff
@@ -55,6 +55,59 @@ class ScalarParamWrapper {
...
+                       "Uint32 value of scalar param");
+        break;
+      case QNN_DATATYPE_SFIXED_POINT_32:
+      case QNN_DATATYPE_INT_32:
+        CHECK_VALUE_EQ(qnn_scalar_.int32Value, other.qnn_scalar_.int32Value,
+                       "Int32 value of scalar param");
+        break;
+      default:
+        QNN_LOG_ERROR(
+            "Unsupported data type for scalar param: input: %#x, golden: %#x",
```
weilhuan-qti:
```suggestion
            "Unsupported data type for comparing scalar param: input: %#x, golden: %#x",
```

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h
```diff
@@ -71,6 +100,42 @@ class AxisScaleOffsetQuantizeParamsWrapper final {
...
+    }
+
+    for (size_t i = 0;
+         i < qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets; i++) {
+      CHECK_VALUE_EQ(
+          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset[i].offset,
+          other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset[i]
+              .offset,
+          ("Offset at ScaleOffset[" + std::to_string(i) + "]").c_str());
+    }
```
weilhuan-qti: Scale and offset can be checked in the same for-loop.

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h
```diff
@@ -71,6 +100,42 @@ class AxisScaleOffsetQuantizeParamsWrapper final {
...
+        other.qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
+        "Number of scaleOffsets of quantize params");
+
+    for (size_t i = 0;
+         i < qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets; i++) {
+      CHECK_VALUE_EQ(
+          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset[i].scale,
+          other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset[i]
+              .scale,
+          ("Scale at ScaleOffset[" + std::to_string(i) + "]").c_str());
```
weilhuan-qti: I think you may pass a invalid pointer because `("Scale at ScaleOffset[" + std::to_string(i) + "]")` is a temporary string.

litert/vendors/qualcomm/core/wrappers/tests/quantize_params_wrapper_test.cc
```diff
@@ -196,5 +216,55 @@ TEST(AxisScaleOffsetQuantizeParamsWrapperTest, QnnConstructorTest) {
...
+  EXPECT_FALSE(wrapper1 == wrapper3);
+  EXPECT_TRUE(wrapper2 == wrapper3);
+}
+
+TEST(BwScaleOffsetQuantizeParamsWrapperTest, EqualityOperatorTest) {
+  std::uint32_t bitwidth1 = 32;
+  float scale1 = 1.5f;
+  std::int32_t zero_point1 = 10;
+  BwScaleOffsetQuantizeParamsWrapper wrapper1(bitwidth1, scale1, zero_point1);
+  std::uint32_t bitwidth2 = 64;
```
weilhuan-qti: Prefer testing the bitwidth which is smaller than 8 bits.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -112,7 +112,45 @@ class TensorWrapper final {
...
+    for (size_t i = 0; i < qnn_tensor_.v2.clientBuf.dataSize; i++) {
+      CHECK_VALUE_EQ(
+          reinterpret_cast<const std::uint8_t*>(
+              qnn_tensor_.v2.clientBuf.data)[i],
+          reinterpret_cast<const std::uint8_t*>(
+              other.qnn_tensor_.v2.clientBuf.data)[i],
+          ("Data of client buffer[" + std::to_string(i) + "]").c_str());
+    }
+    CHECK_VALUE_EQ(qnn_tensor_.v2.isProduced, other.qnn_tensor_.v2.isProduced,
+                   "Tensor.isProduced");
```
weilhuan-qti: Can ignore this check because we don't use it.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -112,7 +112,45 @@ class TensorWrapper final {
...
+                  "Memory type");
+    CHECK_VALUE_EQ(qnn_tensor_.v2.clientBuf.dataSize,
+                   other.qnn_tensor_.v2.clientBuf.dataSize,
+                   "Data size of client buffer");
+    for (size_t i = 0; i < qnn_tensor_.v2.clientBuf.dataSize; i++) {
+      CHECK_VALUE_EQ(
+          reinterpret_cast<const std::uint8_t*>(
+              qnn_tensor_.v2.clientBuf.data)[i],
+          reinterpret_cast<const std::uint8_t*>(
+              other.qnn_tensor_.v2.clientBuf.data)[i],
```
weilhuan-qti: Please left some comment to explain why we use `std::uint8_t` here because we may store different data type.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -112,7 +112,45 @@ class TensorWrapper final {
 
   ~TensorWrapper();
 
-  bool operator==(const TensorWrapper& other) const { return this == &other; }
+  bool operator==(const TensorWrapper& other) const {
+    // Compare the address
+    if (this == &other) {
+      return true;
+    }
+
```
weilhuan-qti, line 116: We should also check quant param.

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -514,5 +514,59 @@ TEST(TensorWrapperTest, QnnTensorNoCopyTest) {
     EXPECT_EQ(ref_data[i], data[i]);
   }
 }
+
+TEST(TensorWrapperTest, EqualityOperatorTest) {
```
weilhuan-qti: Nit: there are many members in `TensorWrapper`, it would be better if we can test various combination of them.

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
```diff
@@ -186,6 +186,18 @@ TEST(ScalarParamWrapperTest, QuantizedInt32ParamTest) {
   EXPECT_EQ(int32_quant_qnn_param.scalarParam.int32Value, value);
 }
 
+TEST(ScalarParamWrapperTest, EqualityOperatorTest) {
```
weilhuan-qti: Nit: use parameterized test to cover more data types.

litert/vendors/qualcomm/core/wrappers/op_wrapper.h
```diff
@@ -26,6 +26,27 @@ class OpWrapper final {
 
   OpWrapper& operator=(const OpWrapper& other);
 
+  bool operator==(const OpWrapper& other) const {
+    CHECK_VALUE_EQ(op_code_, other.op_code_, "Op code in OpWrapper");
+    CHECK_STR_EQ(type_name_, other.type_name_, "Op type name in OpWrapper");
+
+    if (input_tensors_.size() != other.input_tensors_.size()) return false;
+    for (size_t i = 0; i < input_tensors_.size(); ++i) {
+      if (!(input_tensors_[i].get() == other.input_tensors_[i].get())) {
```
weilhuan-qti:
```suggestion
      if (input_tensors_[i].get() != other.input_tensors_[i].get()) {
```

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,131 @@ TEST(OpWrapperTest, ChangeOpName) {
...
+
+  TensorWrapper tensor_wrapper3{"",
+                                QNN_TENSOR_TYPE_APP_WRITE,
+                                QNN_DATATYPE_UFIXED_POINT_8,
+                                QuantizeParamsWrapperVariant(),
+                                dummy_dims3,
+                                static_cast<uint32_t>(data_size3),
+                                data_ptr3,
+                                true};
+
```
weilhuan-qti: If can, please left comments or separate the test cases to indicate what we want to test:
- op type, op code
- input tensors
- output tensors
- scalar param
- tensor param

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
graham0824: Thanks for the reminder! Updated in the latest commit.

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h
graham0824: Thanks for the  comment! Updated in the latest commit.

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.h
graham0824: Yes, it's better to initialize a variable to store the string here, updated in the latest commit.

litert/vendors/qualcomm/core/wrappers/param_wrapper.h
graham0824: Updated in the latest commit.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Sure, change to use static_cast and changed to use string_view to avoid the copy of string

graham0824: In the real case, only 1 error message will be printed and then return false, so I prefer to keep these error messages for the clarity, we could talk about it offline.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
graham0824: Sure, left this now, we can talk about it offline:

- _Since the clientBuf may store different data types (e.g., float, int32_t), and type-aware comparison could fail or misinterpret padding/alignment, Compare client buffer contents byte-by-byte using std::uint8_t to ensure we have an exact match_

litert/vendors/qualcomm/core/wrappers/tests/quantize_params_wrapper_test.cc
graham0824: Sure, change to 4 and 2, also update another test which uses 32 bit.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
graham0824: Thanks for the reminder! Added the comparison of quantized params before return true.

litert/vendors/qualcomm/core/wrappers/op_wrapper.h
graham0824: By using != instead of using ==, need to modify the operator != in TensorWrapper, updated the operator != and use != here in the latest commit, thanks!

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
graham0824: Added some of them, we could add more if needed

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
graham0824: Updated all tests code in parameterized test and cover more options. Updated in the latest commit.

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
graham0824: Added part of them, we can add more if needed

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -20,9 +20,53 @@
 
 #include "absl/types/span.h"  // from @com_google_absl
 #include "litert/vendors/qualcomm/core/schema/soc_table.h"
+#include "litert/vendors/qualcomm/core/utils/log.h"
 #include "QnnInterface.h"  // from @qairt
 #include "QnnTypes.h"  // from @qairt
 namespace qnn {
+namespace {
+
+#define CHECK_STR_EQ(INPUT, GOLDEN, MSG)                                       \
```
ChunHsuehLee: NIT:
Maybe we can also consider inline or lambda function.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -112,9 +112,58 @@ class TensorWrapper final {
...
+                     error_message);
+    }
 
-  bool operator!=(const TensorWrapper& other) const { return this != &other; }
+    // Compare quantize params
+    if (!(quantize_params_ == other.quantize_params_)) {
+      return false;
+    }
+
+    return true;
```
ChunHsuehLee:
```suggestion
    return quantize_params_ == other.quantize_params_;
```

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
```diff
@@ -112,9 +112,58 @@ class TensorWrapper final {
...
+                   "Data format");
+    CHECK_TYPE_EQ(qnn_tensor_.v2.dataType, other.qnn_tensor_.v2.dataType,
+                  "Data type");
+    CHECK_VALUE_EQ(qnn_tensor_.v2.rank, other.qnn_tensor_.v2.rank, "Rank");
+    for (size_t i = 0; i < qnn_tensor_.v2.rank; i++) {
+      std::string error_message =
+          "Dimension of rank [" + std::to_string(i) + "]";
+      CHECK_VALUE_EQ(qnn_tensor_.v2.dimensions[i],
+                     other.qnn_tensor_.v2.dimensions[i], error_message);
+    }
```
ChunHsuehLee: NIT:
Not sure if adding another inline function for checking a span will benefit here and other place like axis quant.

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
```diff
@@ -229,5 +259,23 @@ TEST(ParamWrapperTest, TensorParamTest) {
...
+                            static_cast<uint32_t>(data.size()),
+                            data.data(),
+                            true};
+
+  qnn::TensorParamWrapper param1{"tensor_param", tensor};
+  qnn::TensorParamWrapper param2{"tensor_param", tensor};
+
+  EXPECT_TRUE(param1 == param2);
+}
+
```
ChunHsuehLee, line 310: Do we need to check for expected false cases. Like create different tensor for the 2 params.

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -540,5 +540,101 @@ TEST(TensorWrapperDatatypeTest, SFIXED_POINT_4) {
...
+    std::tuple<Qnn_TensorType_t, Qnn_TensorType_t, Qnn_DataType_t,
+               Qnn_DataType_t, std::vector<uint32_t>, std::vector<uint32_t>,
+               bool>;
+class TensorWrapperEqualTest
+    : public ::testing::TestWithParam<TensorEqualParams> {};
+TEST_P(TensorWrapperEqualTest, EqualityOperator) {
+  auto [tensor_type1, tensor_type2, data_tensor_type1, data_tensor_type2, dims1,
+        dims2, tweak_data] = GetParam();
+
+  auto makeData = [](const std::vector<uint32_t>& dims, uint8_t fill) {
```
ChunHsuehLee:
```suggestion
  auto make_data = [](const std::vector<uint32_t>& dims, uint8_t fill) {
```

litert/vendors/qualcomm/core/wrappers/tests/quantize_params_wrapper_test.cc
```diff
@@ -40,6 +40,13 @@ TEST(UndefinedQuantizeParamsWrapperTest, MoveConstructorTest) {
   EXPECT_EQ(dst.quantizationEncoding, QNN_QUANTIZATION_ENCODING_UNDEFINED);
 }
 
+TEST(UndefinedQuantizeParamsWrapperTest, EqualityOperatorTest) {
+  UndefinedQuantizeParamsWrapper wrapper1;
+  UndefinedQuantizeParamsWrapper wrapper2(wrapper1);
```
weilhuan-qti: I think for `EqualityOperatorTest`, we can just use the same type of constructor (default constructor in this case) to reduce the variable in unit test.

```suggestion
  UndefinedQuantizeParamsWrapper wrapper1;
  UndefinedQuantizeParamsWrapper wrapper2;
```

litert/vendors/qualcomm/core/wrappers/tests/quantize_params_wrapper_test.cc
```diff
@@ -87,6 +94,43 @@ TEST(ScaleOffsetQuantizeParamsWrapperTest, GetterTest) {
...
+
+using ScaleOffsetParamsWithType = std::tuple<float, int32_t, float, int32_t>;
+class ScaleOffsetEqualTestWithType
+    : public ::testing::TestWithParam<ScaleOffsetParamsWithType> {};
+TEST_P(ScaleOffsetEqualTestWithType, EqualityOperator) {
+  auto [scale1, offset1, scale2, offset2] = GetParam();
+
+  qnn::ScaleOffsetQuantizeParamsWrapper wrapper1(scale1, offset1);
+  qnn::ScaleOffsetQuantizeParamsWrapper wrapper2(scale2, offset2);
+  qnn::ScaleOffsetQuantizeParamsWrapper wrapper3(wrapper2);
```
weilhuan-qti: Ditto, use the same constructor.

```suggestion
  qnn::ScaleOffsetQuantizeParamsWrapper wrapper3(scale2, offset2);
```

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
weilhuan-qti: Agree with this.

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
```diff
@@ -186,6 +186,36 @@ TEST(ScalarParamWrapperTest, QuantizedInt32ParamTest) {
   EXPECT_EQ(int32_quant_qnn_param.scalarParam.int32Value, value);
 }
 
+using ScalarParams = std::tuple<Qnn_DataType_t, bool>;
+class ScalarParamEqualTest : public ::testing::TestWithParam<ScalarParams> {};
+TEST_P(ScalarParamEqualTest, EqualityOperator) {
+  auto [data_type, expected_equal] = GetParam();
+
+  if (data_type == QNN_DATATYPE_INT_32) {
```
weilhuan-qti: 1. When using parameterized test, we should prevent such check with different behavior. Otherwise, we can manually split them into simple test.
2. Nit: when using parameterized test, we had better generate all types used in scalar param.
3. I think you can use `TYPED_TEST_SUITE` instead, with a suitable value as scalar for all types, e.g. 0 and 1.

litert/vendors/qualcomm/core/utils/miscs.h
weilhuan-qti: We can discuss about this, I think we can implement a simple version of equity operator which only check value without printing log. And we can add log for debug if we encounter use case in the future.

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,172 @@ TEST(OpWrapperTest, ChangeOpName) {
...
 
+using OpEqualParams = std::tuple<const char*, const char*, QnnOpCode, QnnOpCode,
+                                 std::vector<uint32_t>, std::vector<uint32_t>,
+                                 bool, bool, bool, bool, bool>;
+class OpWrapperEqualParamTest : public ::testing::TestWithParam<OpEqualParams> {
+};
+TEST_P(OpWrapperEqualParamTest, EqualityOperator) {
+  auto [type1, type2, code1, code2, dims1, dims2, differ_scalar_value,
+        differ_tensor_param, differ_input_size, differ_input_data,
+        expected_equal] = GetParam();
```
weilhuan-qti: I think we can separate them into two test, one is for the equal op and another is for the unequal op. Then we can make the param easier to understand.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Discussed offline, change to inline function and not print log.

litert/vendors/qualcomm/core/wrappers/tensor_wrapper.h
graham0824: Good idea, added a CheckSpanEq() in the miscs.h and update the usage in tensor_wrapper & quantize_params_wrapper

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
graham0824: Sure, updated with separate tests

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
graham0824: Thanks for the suggestion, I think the point 1 is very valuable as we're using parameterized test. As we discussed offline, updated with ` TYPED_TEST_SUITE` for this test

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -20,9 +20,36 @@
 
 #include "absl/types/span.h"  // from @com_google_absl
 #include "litert/vendors/qualcomm/core/schema/soc_table.h"
+#include "litert/vendors/qualcomm/core/utils/log.h"
```
ChunHsuehLee:
```suggestion
```

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
```diff
@@ -21,6 +21,18 @@ UndefinedQuantizeParamsWrapper::UndefinedQuantizeParamsWrapper(
...
+bool UndefinedQuantizeParamsWrapper::operator==(
+    const UndefinedQuantizeParamsWrapper& other) const {
+  if (qnn_quantize_param_.encodingDefinition !=
+      other.qnn_quantize_param_.encodingDefinition)
+    return false;
+  if (qnn_quantize_param_.quantizationEncoding !=
+      other.qnn_quantize_param_.quantizationEncoding)
+    return false;
+
+  return true;
```
ChunHsuehLee:
```suggestion
  return qnn_quantize_param_.encodingDefinition ==
      other.qnn_quantize_param_.encodingDefinition && qnn_quantize_param_.quantizationEncoding ==
      other.qnn_quantize_param_.quantizationEncoding;
```

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
```diff
@@ -48,6 +60,23 @@ ScaleOffsetQuantizeParamsWrapper::ScaleOffsetQuantizeParamsWrapper(
...
+      other.qnn_quantize_param_.quantizationEncoding)
+    return false;
+  if (qnn_quantize_param_.scaleOffsetEncoding.scale !=
+      other.qnn_quantize_param_.scaleOffsetEncoding.scale)
+    return false;
+  if (qnn_quantize_param_.scaleOffsetEncoding.offset !=
+      other.qnn_quantize_param_.scaleOffsetEncoding.offset)
+    return false;
+
+  return true;
```
ChunHsuehLee:
```suggestion
    const ScaleOffsetQuantizeParamsWrapper& other) const {
  return qnn_quantize_param_.encodingDefinition ==
      other.qnn_quantize_param_.encodingDefinition && 
  qnn_quantize_param_.quantizationEncoding ==
      other.qnn_quantize_param_.quantizationEncoding &&
  qnn_quantize_param_.scaleOffsetEncoding.scale ==
      other.qnn_quantize_param_.scaleOffsetEncoding.scale &&
  qnn_quantize_param_.scaleOffsetEncoding.offset ==
      other.qnn_quantize_param_.scaleOffsetEncoding.offset;
```

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
```diff
@@ -152,6 +207,27 @@ BwScaleOffsetQuantizeParamsWrapper::BwScaleOffsetQuantizeParamsWrapper(
...
+      other.qnn_quantize_param_.bwScaleOffsetEncoding.bitwidth)
+    return false;
+  if (qnn_quantize_param_.bwScaleOffsetEncoding.scale !=
+      other.qnn_quantize_param_.bwScaleOffsetEncoding.scale)
+    return false;
+  if (qnn_quantize_param_.bwScaleOffsetEncoding.offset !=
+      other.qnn_quantize_param_.bwScaleOffsetEncoding.offset)
+    return false;
+
+  return true;
```
ChunHsuehLee: ditto

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
```diff
@@ -105,6 +134,32 @@ AxisScaleOffsetQuantizeParamsWrapper::AxisScaleOffsetQuantizeParamsWrapper(
...
+  if (qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets !=
+      other.qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets)
+    return false;
+
+  if (!CheckSpanEq(
+          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+          other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+          qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
+          [](const Qnn_ScaleOffset_t& a, const Qnn_ScaleOffset_t& b) {
+            return a.scale == b.scale && a.offset == b.offset;
```
ChunHsuehLee: For floating point comparison, should we consider a tolerance? https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon.html

graham0824: Good question, I thought it was for comparing between the original param and copied param, so we could compare with == directly. But if we would like to use the comparison function for accuracy check, then we should consider comparing them with a tolerance. Let's discuss it offline.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -20,9 +20,36 @@
...
+      (input && golden && std::strcmp(input, golden) != 0)) {
+    return false;
+  }
+
+  return true;
+}
+
+template <typename T, typename Comparison = std::equal_to<T>>
+inline bool CheckSpanEq(const T* input, const T* golden, size_t size,
+                        Comparison compare = Comparison()) {
```
weilhuan-qti: 1. We can check input data size and output data size here.
2. If we call it `CheckSpan`, maybe we can pass span as parameter.
3. We call it `CheckSpanEq`, but the comparator can be changed, it is a little confusing.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -84,5 +111,15 @@ inline const QNN_INTERFACE_VER_TYPE* ResolveQnnApi(
...
 
 std::optional<::qnn::SocInfo> FindSocModel(std::string_view soc_model_name);
+
+inline bool CompareConfig(const Qnn_OpConfigV1_t& input,
+                          const Qnn_OpConfigV1_t& golden) {
+  if (!CheckStrEq(input.packageName, golden.packageName)) return false;
+  if (!CheckStrEq(input.typeName, golden.typeName)) return false;
+  if (input.numOfParams != golden.numOfParams) return false;
+  if (input.numOfInputs != golden.numOfInputs) return false;
+  if (input.numOfOutputs != golden.numOfOutputs) return false;
```
weilhuan-qti: Forget to check `params` `inputTensors` and `outputTensors`.

litert/vendors/qualcomm/core/wrappers/param_wrapper.h
```diff
@@ -66,6 +68,10 @@ class TensorParamWrapper {
  public:
   explicit TensorParamWrapper(const char* name, const TensorWrapper& tensor);
 
+  bool operator==(const TensorParamWrapper& other) const {
+    return tensor_ == other.tensor_;
+  }
```
weilhuan-qti: Should we move this implementation into .cc file? leave it in header file is fine because it is simple.

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,208 @@ TEST(OpWrapperTest, ChangeOpName) {
...
+TensorWrapper CreateTensor(const std::vector<uint32_t>& dims, uint8_t val) {
+  std::vector<uint8_t> data(
+      std::accumulate(dims.begin(), dims.end(), 1u, std::multiplies<>()), val);
+  return TensorWrapper("", QNN_TENSOR_TYPE_APP_WRITE,
+                       QNN_DATATYPE_UFIXED_POINT_8,
+                       QuantizeParamsWrapperVariant(), dims,
+                       static_cast<uint32_t>(data.size()), data.data(), true);
+}
+
+TEST(OpWrapperTest, EqualityOperatorTypeNameTest) {
```
weilhuan-qti: Group these tests by:
```suggestion
TEST(OpWrapperEqualityOperatorTest, TypeName) {
```

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,208 @@ TEST(OpWrapperTest, ChangeOpName) {
...
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, 1);
+  auto output = input;
+
+  qnn::OpWrapper op1{"op", "TYPE_A", QnnOpCode::kUnknown};
+  op1.AddInputTensor(input);
+  op1.AddOutputTensor(output);
+  op1.AddScalarParam("param", static_cast<uint8_t>(255), false);
```
weilhuan-qti:
```suggestion
  op1.AddScalarParam<std::uint8_t>("param", 255, false);
```

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -20,9 +20,36 @@
...
+namespace {
+
+inline bool CheckStrEq(const char* input, const char* golden) {
+  if ((input == nullptr && golden != nullptr) ||
+      (input != nullptr && golden == nullptr) ||
+      (input && golden && std::strcmp(input, golden) != 0)) {
+    return false;
+  }
+
+  return true;
```
KelvinYang0320: This may help enhance readability.
```suggestion
  if (input == nullptr && golden == nullptr) return true;
  if (input == nullptr || golden == nullptr) return false;
  return std::strcmp(input, golden) == 0;
```

litert/vendors/qualcomm/core/utils/miscs.h
KelvinYang0320: > 2. If we call it CheckSpan, maybe we can pass span as parameter.

If the span is not being passed as a parameter, I recommend updating the function name to better reflect its purpose.

litert/vendors/qualcomm/core/utils/miscs.h
KelvinYang0320: Would it be possible to separate this function into another PR that includes the actual use case?

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
graham0824: Let's keep it in this PR, we could add a new function to compare with a tolerance when we have a real use case.

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Discussed offline, replace with std::equal

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
```diff
@@ -105,6 +125,33 @@ AxisScaleOffsetQuantizeParamsWrapper::AxisScaleOffsetQuantizeParamsWrapper(
...
+
+  if (!std::equal(
+          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset +
+              qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
+          other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+          [](const Qnn_ScaleOffset_t& a, const Qnn_ScaleOffset_t& b) {
+            return a.scale == b.scale && a.offset == b.offset;
+          }))
+    return false;
```
ChunHsuehLee: NIT: 
```suggestion
  return std::equal(
          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
          qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset +
              qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
          other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
          [](const Qnn_ScaleOffset_t& a, const Qnn_ScaleOffset_t& b) {
            return a.scale == b.scale && a.offset == b.offset;
          });
```

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,208 @@ TEST(OpWrapperTest, ChangeOpName) {
...
 }
 
+TensorWrapper CreateTensor(const std::vector<uint32_t>& dims, uint8_t val) {
+  std::vector<uint8_t> data(
+      std::accumulate(dims.begin(), dims.end(), 1u, std::multiplies<>()), val);
+  return TensorWrapper("", QNN_TENSOR_TYPE_APP_WRITE,
+                       QNN_DATATYPE_UFIXED_POINT_8,
+                       QuantizeParamsWrapperVariant(), dims,
+                       static_cast<uint32_t>(data.size()), data.data(), true);
+}
```
ChunHsuehLee: NIT:
Maybe we can reuse
https://github.com/google-ai-edge/LiteRT/blob/main/litert/vendors/qualcomm/core/tensor_pool.h#L55

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
graham0824: Thanks for the reminder! Also updated for "BwAxisScaleOffsetQuantizeParamsWrapper"

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
graham0824: The function you pointed out in tensor_pool is for creating static tensor which can't be used for input/output tensor creation. For other functions in tensor_pool , the function for creating native tensor didn't accept a value to fill.

From my view, I would like to keep it here and not create a new function in tensor_pool for testing purpose. We could discuss offline if we still nee this, thanks!

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
```diff
@@ -34,6 +35,30 @@ OpWrapper& OpWrapper::operator=(const OpWrapper& other) {
...
+    return false;
+  }
+
+  if (output_tensors_.size() != other.output_tensors_.size()) return false;
+  if (!std::equal(
+          output_tensors_.begin(), output_tensors_.end(),
+          other.output_tensors_.begin(),
+          [](const auto& a, const auto& b) { return a.get() == b.get(); })) {
+    return false;
+  }
```
KelvinYang0320: `std::equal` can verify whether std::distance(first1, last1) and std::distance(first2, last2) are equal.
Could we consider updating the implementation as follows?
```suggestion
  if (!std::equal(
          input_tensors_.begin(), input_tensors_.end(),
          other.input_tensors_.begin(), other.input_tensors_.end()
          [](const auto& a, const auto& b) { return a.get() == b.get(); })) {
    return false;
  }

  if (!std::equal(
          output_tensors_.begin(), output_tensors_.end(),
          other.output_tensors_.begin(), other.output_tensors_.end(),
          [](const auto& a, const auto& b) { return a.get() == b.get(); })) {
    return false;
  }
```

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,208 @@ TEST(OpWrapperTest, ChangeOpName) {
   EXPECT_STREQ(op_wrapper_1.GetOpConfig().v1.name, "namespace/name_new");
 }
 
+TensorWrapper CreateTensor(const std::vector<uint32_t>& dims, uint8_t val) {
+  std::vector<uint8_t> data(
+      std::accumulate(dims.begin(), dims.end(), 1u, std::multiplies<>()), val);
+  return TensorWrapper("", QNN_TENSOR_TYPE_APP_WRITE,
```
KelvinYang0320: I guess you want to test QNN_TENSOR_TYPE_STATIC instead of QNN_TENSOR_TYPE_APP_WRITE.
```suggestion
  return TensorWrapper("", QNN_TENSOR_TYPE_STATIC,
```

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,208 @@ TEST(OpWrapperTest, ChangeOpName) {
   EXPECT_STREQ(op_wrapper_1.GetOpConfig().v1.name, "namespace/name_new");
 }
 
+TensorWrapper CreateTensor(const std::vector<uint32_t>& dims, uint8_t val) {
```
KelvinYang0320:
```suggestion
TensorWrapper CreateStaticTensor(const std::vector<uint32_t>& dims, uint8_t val) {
```

litert/vendors/qualcomm/core/wrappers/tests/param_wrapper_test.cc
```diff
@@ -229,5 +278,35 @@ TEST(ParamWrapperTest, TensorParamTest) {
...
+                             data1.data(),
+                             true};
+  qnn::TensorWrapper tensor2{"",
+                             QNN_TENSOR_TYPE_APP_WRITE,
+                             QNN_DATATYPE_UFIXED_POINT_8,
+                             QuantizeParamsWrapperVariant(),
+                             {1, 1, 3},
+                             static_cast<uint32_t>(data2.size()),
+                             data2.data(),
+                             true};
```
KelvinYang0320, line 301: Could you change to QNN_TENSOR_TYPE_STATIC?

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -568,5 +568,101 @@ TEST(TensorWrapperDatatypeTest, SFIXED_POINT_2) {
...
+               bool>;
+class TensorWrapperEqualTest
+    : public ::testing::TestWithParam<TensorEqualParams> {};
+TEST_P(TensorWrapperEqualTest, EqualityOperator) {
+  auto [tensor_type1, tensor_type2, data_tensor_type1, data_tensor_type2, dims1,
+        dims2, tweak_data] = GetParam();
+
+  auto make_data = [](const std::vector<uint32_t>& dims, uint8_t fill) {
+    uint32_t count = 1;
+    for (auto dim : dims) count *= dim;
```
KelvinYang0320: Nit:
```suggestion
    uint32_t count = std::accumulate(dims.begin(), dims.end(), 1u, std::multiplies<>());
```

litert/vendors/qualcomm/core/wrappers/tests/tensor_wrapper_test.cc
```diff
@@ -568,5 +568,101 @@ TEST(TensorWrapperDatatypeTest, SFIXED_POINT_2) {
...
+                             TensorEqualParams{QNN_TENSOR_TYPE_APP_WRITE,
+                                               QNN_TENSOR_TYPE_APP_WRITE,
+                                               QNN_DATATYPE_UFIXED_POINT_8,
+                                               QNN_DATATYPE_UFIXED_POINT_8,
+                                               {1, 1, 3},
+                                               {1, 1, 3},
+                                               false},
+                             // Tensor type
+                             TensorEqualParams{QNN_TENSOR_TYPE_APP_WRITE,
+                                               QNN_TENSOR_TYPE_APP_READ,
```
KelvinYang0320: ditto
Maybe you can choose QNN_TENSOR_TYPE_STATIC and QNN_TENSOR_TYPE_UPDATEABLE_STATIC?

litert/vendors/qualcomm/core/wrappers/op_wrapper.cc
graham0824: Thanks for the comment! This is a good optimization we should apply, updated in the latest commit.

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
graham0824: For input/output tensor, we should use APP_WRITE. For tensor param, we should use STATIC. Let me update this function to allow using different type and use STATIC for `AddTensorParam` instead of using `output` as the tensor param.

litert/vendors/qualcomm/core/wrappers/quantize_params_wrapper.cc
```diff
@@ -105,6 +125,30 @@ AxisScaleOffsetQuantizeParamsWrapper::AxisScaleOffsetQuantizeParamsWrapper(
...
+  if (qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets !=
+      other.qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets)
+    return false;
+
+  return std::equal(
+      qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
+      qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset +
+          qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
+      other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
```
weilhuan-qti: Also need to specify the range for `other`.
```suggestion
      other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset,
      other.qnn_quantize_param_.axisScaleOffsetEncoding.scaleOffset +
      other.qnn_quantize_param_.axisScaleOffsetEncoding.numScaleOffsets,
```

litert/vendors/qualcomm/core/wrappers/tests/op_wrapper_test.cc
```diff
@@ -209,5 +209,215 @@ TEST(OpWrapperTest, ChangeOpName) {
...
+      std::accumulate(dims.begin(), dims.end(), 1u, std::multiplies<>()), val);
+  return TensorWrapper("", type, QNN_DATATYPE_UFIXED_POINT_8,
+                       QuantizeParamsWrapperVariant(), dims,
+                       static_cast<uint32_t>(data.size()), data.data(), true);
+}
+
+TEST(OpWrapperEqualityOperatorTest, TypeName) {
+  std::vector<uint32_t> dims = {1, 1, 3};
+  auto input = CreateTensor(dims, 1);
+  auto output = input;
```
KelvinYang0320, line 228: I think that assigning `QNN_TENSOR_TYPE_APP_WRITE` to both the input and output tensors may not be valid in QNN.

To better align with our use case, could we update the configuration so that in[0] is set to `QNN_TENSOR_TYPE_NATIVE`, in[1] to `QNN_TENSOR_TYPE_STATIC`, and out[0] to `QNN_TENSOR_TYPE_NATIVE`?

Additionally, you may consider updating the `CreateTensor`:
```C++
TensorWrapper CreateTensor(const std::vector<uint32_t>& dims, std::optional<uint8_t> val, Qnn_TensorType_t type) 
```
* Use `std::optional` to indicate the presence or absence of a value.
  * No val in QNN_TENSOR_TYPE_NATIVE.
* Removing the default `Qnn_TensorType_t` could improve the readability of this test.

graham0824: Discussed offline, use NATIVE instead of APP_WRITE for input/output, thanks for the reminder!

KelvinYang0320 (approved):

YunanAZ (approved):

PR #5000, opened 2025-12-17
Openvino improvements

akodanka:
Following changes are introduced as part of this PR:
- Additional ops support for intel openvino
- Assign default names of tensors with empty name
- Align the default element type to OV 2026.0
- Add NPU_TURBO config for improving inference.

Tests:
Setup the OPENVINO_NATIVE_DIR with the required libs(including NPU dependencies).
Uncommented https://github.com/google-ai-edge/LiteRT/blob/b6a9f400443fb29e7c878ffb617ea0d65464c555/litert/integration_test/litert_device_common.bzl#L231 to use the npu library.
Ran the ATS test using "bazel run  -c opt --config=android_x86_64   //litert/ats:intel_openvino_ats -- --csv=/data/local/tmp/ptl_ats.csv"


MatthiasKreileder (commented):
Thanks for the PR!

For posterity: Please add a small PR description and also how you tested it (I suppose via ATS).

PR #5007, opened 2025-12-17
[NeuroPilot] modify the way for loading bytecode

neuropilot-captain:
due to compiler plugin generates the cache, so regard the bytecode as cache first.


neuropilot-captain:
Hi @turbotoribio  could you please review this PR? thanks!

PR #5025, opened 2025-12-18
Qualcomm AI Engine Direct - Modify CMakeLists for test files.

ChunHsuehLee:
Summary:
- Clean cmake dependencies for tests.
- Add missing dependency for tensor_wrapper_test.

### DIFF
--- a/litert/vendors/qualcomm/tests/CMakeLists.txt
+++ b/litert/vendors/qualcomm/tests/CMakeLists.txt
@@ -68,8 +68,7 @@ target_sources(common_test
 )
 target_link_libraries(common_test
     PRIVATE
-        qnn_compiler_plugin
-        litert_test_common
+        qnn_core
         GTest::gtest_main
 )
 
@@ -83,8 +82,8 @@ target_sources(tensor_pool_test
 
 target_link_libraries(tensor_pool_test
     PRIVATE
-        qnn_compiler_plugin
-        litert_test_common
+        qnn_wrappers
+        qnn_core
         GTest::gtest_main
 )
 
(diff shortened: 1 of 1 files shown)


litert/vendors/qualcomm/tests/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ target_sources(common_test
 )
 target_link_libraries(common_test
     PRIVATE
-        qnn_compiler_plugin
+        qnn_core
         litert_test_common
```
weilhuan-qti: Can we try to remove the dependency of `litert_test_common`? I think we should not use this in core module.

litert/vendors/qualcomm/tests/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ target_sources(common_test
 )
 target_link_libraries(common_test
     PRIVATE
-        qnn_compiler_plugin
+        qnn_core
```
weilhuan-qti, line 71: Nit: currently `qnn_core` contain some files which are independent to others. We can separate them into smaller targets and reduce dependency when including `qnn_core`. But I think we can do it later if we want to refine the build system and file structure of core module.

ChunHsuehLee: Totally agree with this, created a ticket to resolve this issue later. Thanks for the reminder.

litert/vendors/qualcomm/tests/CMakeLists.txt
ChunHsuehLee: Yes, thanks for catching this.

YunanAZ (approved):

PR #5118, opened 2025-12-30
Qualcomm AI Engine Direct - Refactor perf control of HTP/DSP backend.

graham0824:
Summary:
- Enable performance control module for DSP backend.
- Add related tests.

Test Result:
- No performance impact for HTP backend
- Pass JIT flow with performance mode for DSP backend
- Pass unit tests on x86
- Pass unit test on device (v81 & v66)

======================== Test Summary ========================
//litert/c/options:litert_qualcomm_options_test
//litert/c/options:litert_qualcomm_options_test                          PASSED in 0.0s

//litert/tools/flags/vendors:qualcomm_flags_test
//litert/tools/flags/vendors:qualcomm_flags_test                         PASSED in 0.0s

//litert/vendors/qualcomm/core/utils:utils_test
//litert/vendors/qualcomm/core/utils:utils_test                          PASSED in 0.1s

//litert/vendors/qualcomm/core/wrappers/tests:op_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:op_wrapper_test            PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:tensor_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:tensor_wrapper_test        PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:param_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:param_wrapper_test         PASSED in 0.0s

//litert/vendors/qualcomm/core/wrappers/tests:quantize_params_wrapper_test
//litert/vendors/qualcomm/core/wrappers/tests:quantize_params_wrapper_test PASSED in 0.0s

//litert/vendors/qualcomm/core:common_test
//litert/vendors/qualcomm/core:common_test                               PASSED in 0.0s

//litert/vendors/qualcomm/core:tensor_pool_test
//litert/vendors/qualcomm/core:tensor_pool_test                          PASSED in 0.0s

//litert/vendors/qualcomm/core/transformation:all
//litert/vendors/qualcomm/core/transformation:embedding_gemma_test       PASSED in 0.0s
//litert/vendors/qualcomm/core/transformation:graph_to_graph_test        PASSED in 0.0s

//litert/vendors/qualcomm/qnn_backend_test:all
//litert/vendors/qualcomm/qnn_backend_test:qnn_model_test                PASSED in 0.5s

//litert/vendors/qualcomm/qnn_backend_test/builder_test:all
//litert/vendors/qualcomm/qnn_backend_test/builder_test:elementwise_test PASSED in 0.9s
//litert/vendors/qualcomm/qnn_backend_test/builder_test:fully_connected_int2_test PASSED in 0.5s
//litert/vendors/qualcomm/qnn_backend_test/builder_test:relu_test        PASSED in 0.5s
//litert/vendors/qualcomm/qnn_backend_test/builder_test:topk_test        PASSED in 0.5s

//litert/vendors/qualcomm/core/dump:dump_graph_test
//litert/vendors/qualcomm/core/dump:dump_graph_test                      PASSED in 0.0s

//litert/vendors/qualcomm:qnn_manager_test
//litert/vendors/qualcomm:qnn_manager_test                               PASSED in 0.3s

//litert/vendors/qualcomm/core/backends:backend_utils_test
//litert/vendors/qualcomm/core/backends:backend_utils_test               PASSED in 0.0s

//litert/vendors/qualcomm/core/backends:htp_backend_test
//litert/vendors/qualcomm/core/backends:htp_backend_test                 PASSED in 0.1s

//litert/vendors/qualcomm/core/backends:ir_backend_test
//litert/vendors/qualcomm/core/backends:ir_backend_test                  PASSED in 0.0s

//litert/c:litert_op_options_test
//litert/c:litert_op_options_test                                        PASSED in 0.0s

//litert/vendors/qualcomm/compiler:qnn_compiler_plugin_test
//litert/vendors/qualcomm/compiler:qnn_compiler_plugin_test              PASSED in 58.2s

======================== Test Summary ========================
SM8850: //litert/c/options:litert_qualcomm_options_test
[==========] 22 tests from 2 test suites ran. (1 ms total)
[  PASSED  ] 22 tests.

SM8850: //litert/tools/flags/vendors:qualcomm_flags_test
[==========] 12 tests from 8 test suites ran. (3 ms total)
[  PASSED  ] 12 tests.

SM8850: //litert/vendors/qualcomm/core/utils:utils_test
[==========] 13 tests from 3 test suites ran. (17 ms total)
[  PASSED  ] 13 tests.

SM8850: //litert/vendors/qualcomm/core/wrappers/tests:op_wrapper_test
[==========] 20 tests from 2 test suites ran. (2 ms total)
[  PASSED  ] 20 tests.

SM8850: //litert/vendors/qualcomm/core/wrappers/tests:tensor_wrapper_test
[==========] 30 tests from 3 test suites ran. (5 ms total)
[  PASSED  ] 30 tests.

SM8850: //litert/vendors/qualcomm/core/wrappers/tests:param_wrapper_test
[==========] 31 tests from 17 test suites ran. (2 ms total)
[  PASSED  ] 31 tests.

SM8850: //litert/vendors/qualcomm/core/wrappers/tests:quantize_params_wrapper_test
[==========] 64 tests from 9 test suites ran. (4 ms total)
[  PASSED  ] 64 tests.

SM8850: //litert/vendors/qualcomm/core:common_test
[==========] 16 tests from 1 test suite ran. (3 ms total)
[  PASSED  ] 16 tests.

SM8850: //litert/vendors/qualcomm/core:tensor_pool_test
[==========] 17 tests from 1 test suite ran. (2 ms total)
[  PASSED  ] 17 tests.

SM8850: //litert/vendors/qualcomm/core/transformation:graph_to_graph_test
[==========] 8 tests from 4 test suites ran. (14 ms total)
[  PASSED  ] 8 tests.

SM8850: //litert/vendors/qualcomm/core/transformation:embedding_gemma_test
[==========] 1 test from 1 test suite ran. (3 ms total)
[  PASSED  ] 1 test.

SM8850: //litert/vendors/qualcomm/qnn_backend_test:qnn_model_test
[==========] 1 test from 1 test suite ran. (707 ms total)
[  PASSED  ] 1 test.

SM8850: //litert/vendors/qualcomm/qnn_backend_test/builder_test:fully_connected_int2_test
[==========] 1 test from 1 test suite ran. (841 ms total)
[  PASSED  ] 1 test.

SM8850: //litert/vendors/qualcomm/qnn_backend_test/builder_test:relu_test
[==========] 1 test from 1 test suite ran. (788 ms total)
[  PASSED  ] 1 test.

SM8850: //litert/vendors/qualcomm/qnn_backend_test/builder_test:elementwise_test
[==========] 2 tests from 1 test suite ran. (1382 ms total)
[  PASSED  ] 2 tests.

SM8850: //litert/vendors/qualcomm/qnn_backend_test/builder_test:topk_test
[==========] 1 test from 1 test suite ran. (815 ms total)
[  PASSED  ] 1 test.

SM8850: //litert/vendors/qualcomm/core/dump:dump_graph_test
[==========] 5 tests from 1 test suite ran. (3 ms total)
[  PASSED  ] 5 tests.

SM8850: //litert/vendors/qualcomm:qnn_manager_test
[==========] 3 tests from 1 test suite ran. (437 ms total)
[  PASSED  ] 3 tests.

SM8850: //litert/vendors/qualcomm/core/backends:backend_utils_test
[==========] 3 tests from 1 test suite ran. (1 ms total)
[  PASSED  ] 3 tests.

SM8850: //litert/vendors/qualcomm/core/backends:htp_backend_test
[==========] 11 tests from 3 test suites ran. (2215 ms total)
[  PASSED  ] 11 tests.

SM8850: //litert/vendors/qualcomm/core/backends:ir_backend_test
[==========] 2 tests from 1 test suite ran. (37 ms total)
[  PASSED  ] 2 tests.

SM8850: //litert/vendors/qualcomm/core/backends:dsp_backend_test
[==========] 10 tests from 2 test suites ran. (7 ms total)
[  PASSED  ] 0 tests.

SM8850: //litert/vendors/qualcomm/dispatch:_dispatch_api_qualcomm_test
[==========] 4 tests from 1 test suite ran. (483 ms total)
[  PASSED  ] 4 tests.

SM8850: //litert/cc:_litert_compiled_model_qualcomm_test
[==========] 2 tests from 1 test suite ran. (429 ms total)
[  PASSED  ] 2 tests.


litert/vendors/qualcomm/core/backends/dsp_backend.h
```diff
@@ -4,12 +4,17 @@
...
+#include <list>
+#include <memory>
 #include <optional>
+#include <vector>
 
 #include "litert/vendors/qualcomm/core/backends/qnn_backend.h"
 #include "litert/vendors/qualcomm/core/common.h"
 #include "litert/vendors/qualcomm/core/schema/soc_table.h"
 #include "DSP/QnnDspCommon.h"  // from @qairt
+#include "QnnDevice.h"  // from @qairt
```
weilhuan-qti: If these header are used for `DspPerfControl`, we can move them to .cc because we use pimpl.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,25 +3,208 @@
...
 #include "QnnBackend.h"  // from @qairt
 #include "QnnInterface.h"  // from @qairt
+#include "litert/vendors/qualcomm/core/utils/miscs.h"
 
 namespace qnn {
 
+// DSP PERF CONTROL /////////////////////////////////////////////////////////
+class DspBackend::DspPerfControl {
+ public:
+  static constexpr int kNumPowerConfigs = 6;
```
weilhuan-qti: Can be private.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,25 +3,208 @@
...
+                    ::qnn::PowerConfig::kSleepHighLatency,
+                    DCVS_VOLTAGE_VCORNER_SVS2, DCVS_VOLTAGE_VCORNER_SVS,
+                    DCVS_VOLTAGE_VCORNER_SVS);
+  }
+
+  const QNN_INTERFACE_VER_TYPE* api_{nullptr};
+  std::uint32_t power_config_id_{0};
+  QnnDevice_Infrastructure_t dsp_perf_infra_{nullptr};
+  std::array<QnnDspPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>
+      perf_power_configs_;
```
weilhuan-qti: Consistent with `down_vote_power_configs_`.
```suggestion
      up_vote_power_configs_;
```

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,25 +3,208 @@
...
+  std::array<QnnDspPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>
+      down_vote_power_configs_;
+  std::array<const QnnDspPerfInfrastructure_PowerConfig_t*,
+             kNumPowerConfigs + 1>
+      perf_power_configs_ptr_;
```
weilhuan-qti:
```suggestion
      up_vote_power_configs_ptr_;
```

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,25 +3,208 @@
...
+  void InitUpVotePowerConfigs(DspPerformanceMode perf_mode) {
+    // 0: sleep_latency, 1: dcvs_vcorner_min, 2: bus_vcorner_target,
+    // 3: bus_vcorner_max, 4: dcvs_enable, 5: dcvs_power_mode
+    perf_power_configs_ = {
+        {{QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_SLEEP_LATENCY},
+         {QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER},
+         {QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER},
+         {QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER},
+         {QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_ENABLE},
+         {QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_POWER_MODE}}};
```
weilhuan-qti: For a `QnnDspPerfInfrastructure_PowerConfig_t`, it is like a pair of type and value. And I prefer set type and value at the same time (same place). And I also prefer set all of them inside one function.

```cpp
static void SetPowerConfigs(
      std::array<QnnDspPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>&
          power_configs,
      uint32_t sleep_latency, QnnDspPerfInfrastructure_VoltageCorner_t voltage_min,
      QnnDspPerfInfrastructure_VoltageCorner_t voltage_target,
      QnnDspPerfInfrastructure_VoltageCorner_t voltage_max, QnnDspPerfInfrastructure_DcvsEnable_t dcvs_enable, QnnDspPerfInfrastructure_PowerMode_t power_mode) {
    constexpr size_t kSleepLatency = 0;
    constexpr size_t kDcvsVcornerMin = 1;
    constexpr size_t kDcvsVcornerTarget = 2;
    constexpr size_t kDcvsVcornerMax = 3;
    constexpr size_t kDcvsEnable = 4;
    constexpr size_t kDcvsPowerMode = 5;
    power_configs[kSleepLatency].config = QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_SLEEP_LATENCY;
    power_configs[kSleepLatency].sleepLatencyConfig = sleep_latency;
    power_configs[kDcvsVcornerMin].config = QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER;
    power_configs[kDcvsVcornerMin].dcvsVoltageCornerMinConfig = min;
    power_configs[kDcvsVcornerTarget].config = QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER;
    power_configs[kDcvsVcornerTarget].busVoltageCornerTargetConfig = target;
    power_configs[kDcvsVcornerMax].config = QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER;
    power_configs[kDcvsVcornerMax].busVoltageCornerMaxConfig = max;
    power_configs[kDcvsEnable].config = QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_ENABLE;
    power_configs[kDcvsEnable].dcvsEnableConfig= dcvs_enable;
    power_configs[kDcvsPowerMode].config = QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_POWER_MODE
    power_configs[kDcvsPowerMode].dcvsPowerModeConfig= power_mode;
  }
```

I know your implementation is shortest in the current use case. But it also make `SetPowerConfigs` become a restricted function, before we call it we need to initialize part of these configs.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,25 +3,208 @@
...
+
+  bool Init(DspPerformanceMode performance_mode) {
+    QnnDevice_Infrastructure_t local_dsp_perf_infra = nullptr;
+    Qnn_ErrorHandle_t error = QNN_SUCCESS;
+
+    if (error = api_->deviceGetInfrastructure(&local_dsp_perf_infra);
+        error != QNN_SUCCESS) {
+      QNN_LOG_ERROR(
+          "DSP backend unable to create device infrastructure. Error %d",
+          QNN_GET_ERROR_CODE(error));
```
weilhuan-qti, line 45: What is `QNN_GET_ERROR_CODE`...

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,25 +3,208 @@
...
+             kNumPowerConfigs + 1>
+      down_vote_power_configs_ptr_;
+};
+
+// DSP BACKEND /////////////////////////////////////////////////////////
 DspBackend::DspBackend(const QNN_INTERFACE_VER_TYPE* qnn_api)
     : QnnBackend(qnn_api) {}
 
-DspBackend::~DspBackend() = default;
+DspBackend::~DspBackend() {}
```
weilhuan-qti: Can we mark it `default`?

litert/vendors/qualcomm/core/common.h
```diff
@@ -52,6 +52,24 @@ enum class HtpPerformanceMode {
...
+  kBurst = 2,
+  kHighPerformance = 3,
+  kPowerSaver = 4,
+  kLowPowerSaver = 5,
+  kHighPowerSaver = 6,
+  kLowBalanced = 7,
+  kBalanced = 8,
+};
+
+enum PerformanceModeVoteType {
```
weilhuan-qti: Please remove it if we don't use it now.

litert/vendors/qualcomm/core/common.h
```diff
@@ -66,6 +84,20 @@ enum class GraphPriority {
   kHigh = 4,
 };
 
+struct PowerConfig {
```
weilhuan-qti: I am thinking create a `backend_utils.h` to put `PowerConfig` and `ObtainNullTermPtrArray` because they are only used for backend now.

litert/vendors/qualcomm/core/utils/miscs.h
```diff
@@ -63,6 +63,17 @@ void DequantizeInto(const absl::Span<const T>& in, const float scale,
   }
 }
 
+template <typename T, std::size_t N>
+std::array<std::add_pointer_t<std::add_const_t<T>>, N + 1>
+ObtainNullTermPtrArray(const std::array<T, N>& arr) {
```
weilhuan-qti: Can we change the interface like:
```suggestion
void SetNullTermPtrArray(const std::array<T, N>& src, std::array<std::add_pointer_t<std::add_const_t<T>>, N + 1> &dst) {
```

litert/vendors/qualcomm/core/backends/htp_backend.h
```diff
@@ -67,13 +67,13 @@ class HtpBackend : public QnnBackend {
   QnnDevicePlatformInfo CreateDevicePlatformInfo();
 
   // TODO: Remove this after user can pass graph options to QNN
-  ::qnn::SocInfo soc_info_ = ::qnn::kSocInfos[7];  // V75
+  ::qnn::SocInfo soc_info_ = kSocInfos[7];  // V75
+  QnnDevicePlatformInfo qnn_device_platform_info_;
```
weilhuan-qti, line 76: If we change the member is because the destruct order, we had better leave some comments to explain.

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -5,25 +5,379 @@
...
+      std::array<QnnHtpPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>>
+      perf_power_configs_;
+  std::optional<
+      std::array<QnnHtpPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>>
+      down_vote_power_configs_;
+  std::array<QnnHtpPerfInfrastructure_PowerConfig_t, kNumRpcPollingPowerConfigs>
+      rpc_power_configs_;
+  std::array<const QnnHtpPerfInfrastructure_PowerConfig_t*,
+             kNumRpcPollingPowerConfigsPtr>
+      rpc_power_configs_ptr_;
```
weilhuan-qti, line 349:
```suggestion
  std::array<const QnnHtpPerfInfrastructure_PowerConfig_t*,
             kNumRpcPollingPowerConfigsPtr + 1>
      rpc_power_configs_ptr_;
```

litert/vendors/qualcomm/core/backends/dsp_backend.cc
graham0824: Qnn_ErrorHandle_t contains internal tracking bits in the higher order bits. 
QNN_GET_ERROR_CODE is a macro from QnnCommon.h to required to mask them out and extract the actual error code defined in the QNN API spec. We could get a error code like 1000, 1002 after calling this.

litert/vendors/qualcomm/core/utils/miscs.h
ChunHsuehLee:
```suggestion
GetNullTermPtrArray(const std::array<T, N>& arr) {
```

litert/vendors/qualcomm/core/backends/dsp_backend.h
graham0824: Sure, moved to .cc for both HTP and DSP backend.

litert/vendors/qualcomm/core/backends/htp_backend.h
graham0824: Thanks for the reminder! The updated order will ensure the qnn_device_platform_info_ used in device configurations are safe. Let me add a comment for it.

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: The kNumRpcPollingPowerConfigsPtr  is 3 and kNumRpcPollingPowerConfigs is 2. But yes, let me use kNumRpcPollingPowerConfigs +1 to align other configs format

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,222 @@
...
-#include "litert/vendors/qualcomm/core/backends/qnn_backend.h"
-#include "litert/vendors/qualcomm/core/common.h"
 #include "litert/vendors/qualcomm/core/schema/soc_table.h"
+#include "litert/vendors/qualcomm/core/utils/backend_utils.h"
 #include "litert/vendors/qualcomm/core/utils/log.h"
+#include "DSP/QnnDspDevice.h"  // from @qairt
+#include "DSP/QnnDspPerfInfrastructure.h"  // from @qairt
 #include "QnnBackend.h"  // from @qairt
 #include "QnnInterface.h"  // from @qairt
+#include "litert/vendors/qualcomm/core/utils/miscs.h"
```
ChunHsuehLee: NIT:
This should be on line 18.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,222 @@
...
+        break;
+      default:
+        QNN_LOG_INFO("performance profile %d don't need to set power configs",
+                     perf_mode);
+        SetPowerConfigs(up_vote_power_configs_, 0,
+                        (QnnDspPerfInfrastructure_VoltageCorner_t)0,
+                        (QnnDspPerfInfrastructure_VoltageCorner_t)0,
+                        (QnnDspPerfInfrastructure_VoltageCorner_t)0,
+                        ::qnn::PowerConfig::kDcvsDisable,
+                        QNN_DSP_PERF_INFRASTRUCTURE_POWERMODE_PERFORMANCE_MODE);
```
ChunHsuehLee: Can we use default config here with a log like "Unknown perf mode, use default."

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,222 @@
...
+                    QNN_GET_ERROR_CODE(error));
+      return false;
+    }
+
+    // Initialize power configurations.
+    // We need to prepare both:
+    // 1. UpVote config: for entering performance mode.
+    // 2. DownVote config: for resetting/cleanup.
+    InitUpVotePowerConfigs(performance_mode);
+    SetNullTermPtrArray(up_vote_power_configs_, up_vote_power_configs_ptr_);
```
ChunHsuehLee: How about putting `SetNullTermPtrArray` inside the init function above.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,222 @@
...
+
+    InitDownVotePowerConfigs();
+    SetNullTermPtrArray(down_vote_power_configs_, down_vote_power_configs_ptr_);
```
ChunHsuehLee: Ditto

litert/vendors/qualcomm/core/utils/BUILD
```diff
@@ -26,6 +26,12 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "backend_utils",
+    hdrs = ["backend_utils.h"],
+    deps = [],
+)
+
```
ChunHsuehLee: Is this unused?

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,376 @@
...
-#include <optional>
 #include <utility>
 #include <vector>
 
 #include "absl/types/span.h"  // from @com_google_absl
-#include "litert/vendors/qualcomm/core/backends/htp_perf_control.h"
 #include "litert/vendors/qualcomm/core/backends/qnn_backend.h"
 #include "litert/vendors/qualcomm/core/common.h"
 #include "litert/vendors/qualcomm/core/schema/soc_table.h"
+#include "litert/vendors/qualcomm/core/utils/backend_utils.h"
```
ChunHsuehLee: Is this unused?

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,376 @@
...
+      htp_perf_infra_->perfInfra.setPowerConfig(
+          power_config_id_, down_vote_power_configs_ptr_.data());
+    }
+  }
+
+ private:
+  static constexpr int kNumPowerConfigs = 1;
+  static constexpr int kNumRpcPollingPowerConfigs = 2;
+
+  static void SetPowerConfigs(
```
ChunHsuehLee: This function is used only in the class, can we make this non-static?

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,376 @@
...
+    // Initialize power configurations.
+    // We need to prepare both:
+    // 1. UpVote config: for entering performance mode.
+    // 2. DownVote config: for resetting/cleanup.
+    if (!InitUpVotePowerConfigs(performance_mode)) {
+      QNN_LOG_ERROR(
+          "Failed to init perf control, using default performance mode.");
+      return false;
+    }
+    SetNullTermPtrArray(*up_vote_power_configs_, up_vote_power_configs_ptr_);
```
ChunHsuehLee: ditto

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,376 @@
...
+
+    if (!InitDownVotePowerConfigs(performance_mode)) {
+      QNN_LOG_ERROR(
+          "Failed to init perf control, using default performance mode.");
+      return false;
+    }
+    SetNullTermPtrArray(*down_vote_power_configs_,
```
ChunHsuehLee: ditto

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,376 @@
...
+                        down_vote_power_configs_ptr_);
+
+    htp_perf_infra_ = std::move(local_htp_perf_infra);
+
+    return true;
+  }
+
+  bool InitRpcPolling(HtpPerformanceMode performance_mode) {
```
ChunHsuehLee:
```suggestion
  bool SetRpcPolling(HtpPerformanceMode performance_mode) {
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -41,9 +390,7 @@ HtpBackend::HtpBackend(const QNN_INTERFACE_VER_TYPE* qnn_api)
     : QnnBackend(qnn_api),
       qnn_device_platform_info_(nullptr, PlatformInfoDeleter{QnnApi()}) {}
 
-HtpBackend::~HtpBackend() {
-  if (perf_control_) perf_control_->Terminate();
-}
+HtpBackend::~HtpBackend() = default;
```
ChunHsuehLee, line 362: Maybe also explicitly define or delete copy/move constructor/assignment.

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -201,23 +548,29 @@ bool HtpBackend::Init(const Options& options,
...
     } else {
       QnnHtpDevice_Arch_t local_arch =
           local_qnn_device_platform_info->v1.hwDevices->v1.deviceInfoExtension
               ->onChipDevice.arch;
-      if (auto status = perf_control_->Init(local_arch); !status) {
-        QNN_LOG_ERROR(
-            "Failed to init perf control, using default performance mode.");
+      htp_perf_control_ = std::make_unique<HtpPerfControl>(QnnApi());
+      if (!htp_perf_control_->Init(performance_mode)) {
+        QNN_LOG_ERROR("Failed to initialize HTP performance Control!");
```
ChunHsuehLee: NIT: How about less excited. 😆 
```suggestion
        QNN_LOG_ERROR("Failed to initialize HTP performance Control.");
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -201,23 +548,29 @@ bool HtpBackend::Init(const Options& options,
...
         return false;
       }
+
+      if (local_arch >= QNN_HTP_DEVICE_ARCH_V69) {
+        if (!htp_perf_control_->InitRpcPolling(performance_mode)) {
+          QNN_LOG_ERROR("Failed to initialize HTP RPC polling!");
```
ChunHsuehLee: ditto

litert/vendors/qualcomm/core/utils/miscs.h
graham0824: Based on the previous alen's comment, changed to SetNullTermPtrArray() FYI, we could rename to GetNullTermPtrArray  if needed later

litert/vendors/qualcomm/core/utils/BUILD
graham0824: I’ve identified the issue! The file wasn’t uploaded in this PR. I’ll add it in the latest commit. Thanks for pointing it out!

Regarding your question, this header includes the PowerConfig struct and SetNullTermPtrArray() used in dsp_backend.h and htp_backend.h. Since these are only used in the backend .cc files, we moved them from miscs.h to a new file called "backend_utils.h".

Based on the above change, we could stop to include the miscs.h in dsp_backend.cc and htp_backend.cc

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Same as here: https://github.com/google-ai-edge/LiteRT/pull/5118#discussion_r2697397997

litert/vendors/qualcomm/core/backends/dsp_backend.cc
weilhuan-qti: Oh, thanks for explanation. I misunderstood that `QNN_GET_ERROR_CODE` is defined in delegate.

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Thanks for the comment! As we hold a unique pointer in the HtpBackend and DspBackend, we should follow the "Rule of Five" to explicitliy define copy/move constructor/assignment.

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Sure XD, changed from"." to "!" for all logs.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
KelvinYang0320: Could you add `#include "QnnCommon.h"  // from @qairt`?

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,212 @@
...
+    }
+
+    // Initialize power configurations.
+    // We need to prepare both:
+    // 1. UpVote config: for entering performance mode.
+    // 2. DownVote config: for resetting/cleanup.
+    InitUpVotePowerConfigs(performance_mode);
+    InitDownVotePowerConfigs();
+
+    dsp_perf_infra_ = std::move(local_dsp_perf_infra);
```
KelvinYang0320:
```suggestion
    dsp_perf_infra_ = local_dsp_perf_infra;
```
Additionally, could we directly replace `local_dsp_perf_infra` with `dsp_perf_infra_`?

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,212 @@
...
+      case ::qnn::DspPerformanceMode::kBurst:
+        SetPowerConfigs(up_vote_power_configs_,
+                        ::qnn::PowerConfig::kSleepMinLatency,
+                        DCVS_VOLTAGE_VCORNER_MAX_VOLTAGE_CORNER,
+                        DCVS_VOLTAGE_VCORNER_MAX_VOLTAGE_CORNER,
+                        DCVS_VOLTAGE_VCORNER_MAX_VOLTAGE_CORNER,
+                        ::qnn::PowerConfig::kDcvsDisable,
+                        QNN_DSP_PERF_INFRASTRUCTURE_POWERMODE_PERFORMANCE_MODE);
+        break;
+      case ::qnn::DspPerformanceMode::kSustainedHighPerformance:
```
KelvinYang0320: Nit: Add [[fallthrough]];

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,21 +3,212 @@
...
+    power_configs[kDcvsEnable].dcvsEnableConfig = dcvs_enable;
+
+    power_configs[kDcvsPowerMode].config =
+        QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_POWER_MODE;
+    power_configs[kDcvsPowerMode].dcvsPowerModeConfig = power_mode;
+  }
+
+  void InitUpVotePowerConfigs(DspPerformanceMode perf_mode) {
+    switch (perf_mode) {
+      case ::qnn::DspPerformanceMode::kBurst:
```
KelvinYang0320: Could we remove all `::qnn` when these functions are in namespace qnn?
```suggestion
      case DspPerformanceMode::kBurst:
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+      case ::qnn::HtpPerformanceMode::kLowBalanced:
+      case ::qnn::HtpPerformanceMode::kBalanced:
+      case ::qnn::HtpPerformanceMode::kDefault:
+      case ::qnn::HtpPerformanceMode::kExtremePowerSaver:
+        rpc_control_latency.option =
+            QNN_HTP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_RPC_CONTROL_LATENCY;
+        rpc_control_latency.rpcControlLatencyConfig =
+            ::qnn::PowerConfig::kRpcControlLatency;
+        rpc_power_configs_[config_count++] = rpc_control_latency;
+        break;
```
KelvinYang0320, line 324: It appears that, for certain performance levels, only rpc_power_configs_[0] is being populated. We may need to revise the current logic; otherwise, SetNullTermPtrArray could create {configs, unknown configs, nullptr}.
Also, could you add [[fallthrough]]?

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+      return false;
+    }
+
+    // Initialize power configurations.
+    // We need to prepare both:
+    // 1. UpVote config: for entering performance mode.
+    // 2. DownVote config: for resetting/cleanup.
+    if (!InitUpVotePowerConfigs(performance_mode)) {
+      QNN_LOG_ERROR(
+          "Failed to init perf control, using default performance mode.");
```
weilhuan-qti: Do we need to return `bool` from `InitUpVotePowerConfigs`? The only case it will return false is when the `performance_mode` is default, and we can check this before we call `InitUpVotePowerConfigs`. But I think both implementations are fine.
```suggestion
          "Failed to init up vote power config, using default performance mode.");
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+  // Performance control
+  const QNN_INTERFACE_VER_TYPE* api_{nullptr};
+  std::uint32_t power_config_id_{0};
+  QnnDevice_Infrastructure_t htp_perf_infra_{nullptr};
+  std::optional<
+      std::array<QnnHtpPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>>
+      up_vote_power_configs_;
+  std::optional<
+      std::array<QnnHtpPerfInfrastructure_PowerConfig_t, kNumPowerConfigs>>
+      down_vote_power_configs_;
```
KelvinYang0320: Could we remove std::optional?
Or could you clarify the reason to have `std::optional` here? Thank you!

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+    dcvs_config.option = QNN_HTP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_V3;
+    QnnHtpPerfInfrastructure_DcvsV3_t& dcvs_v3 = dcvs_config.dcvsV3Config;
+
+    dcvs_v3.contextId = context_id;
+    dcvs_v3.setSleepDisable = 0;  // false
+    dcvs_v3.sleepDisable = 0;     // false
+    dcvs_v3.setDcvsEnable = 1;    // true
+    dcvs_v3.setSleepLatency = 1;  // true
+    dcvs_v3.setBusParams = 1;     // true
+    dcvs_v3.setCoreParams = 1;    // true
```
KelvinYang0320: Could we remove `// false` and `// true`?

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+      return false;
+    }
+
+    if (!InitDownVotePowerConfigs(performance_mode)) {
+      QNN_LOG_ERROR(
+          "Failed to init perf control, using default performance mode.");
+      return false;
+    }
+
+    htp_perf_infra_ = std::move(local_htp_perf_infra);
```
KelvinYang0320: ditto as DSP comment

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+    dcvs_v3.powerMode = power_mode;
+    dcvs_v3.sleepLatency = sleep_latency;
+
+    dcvs_v3.busVoltageCornerMin = bus_voltage_corner_min;
+    dcvs_v3.busVoltageCornerTarget = bus_voltage_corner_target;
+    dcvs_v3.busVoltageCornerMax = bus_voltage_corner_max;
+
+    dcvs_v3.coreVoltageCornerMin = core_voltage_corner_min;
+    dcvs_v3.coreVoltageCornerTarget = core_voltage_corner_target;
+    dcvs_v3.coreVoltageCornerMax = core_voltage_corner_max;
```
weilhuan-qti, line 140: We can re-organize them to make it more readable.
```suggestion
    dcvs_v3.contextId = context_id;
    dcvs_v3.powerMode = power_mode;
    
    dcvs_v3.setDcvsEnable = 1;    // true
    dcvs_v3.dcvsEnable = dcvs_enable;
    
    dcvs_v3.setSleepDisable = 0;  // false
    dcvs_v3.sleepDisable = 0;     // false

    dcvs_v3.setSleepLatency = 1;  // true
    dcvs_v3.sleepLatency = sleep_latency;
    
    dcvs_v3.setBusParams = 1;     // true
    dcvs_v3.busVoltageCornerMin = bus_voltage_corner_min;
    dcvs_v3.busVoltageCornerTarget = bus_voltage_corner_target;
    dcvs_v3.busVoltageCornerMax = bus_voltage_corner_max;

    dcvs_v3.setCoreParams = 1;    // true
    dcvs_v3.coreVoltageCornerMin = core_voltage_corner_min;
    dcvs_v3.coreVoltageCornerTarget = core_voltage_corner_target;
    dcvs_v3.coreVoltageCornerMax = core_voltage_corner_max;
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+    SetNullTermPtrArray(power_configs, down_vote_power_configs_ptr_);
+
+    return true;
+  }
+
+  bool InitRpcPollingPowerConfig(HtpPerformanceMode performance_mode) {
+    rpc_power_configs_ = {
+        {{QNN_HTP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_UNKNOWN},
+         {QNN_HTP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_UNKNOWN}}};
+    int config_count = 0;
```
KelvinYang0320: Nit:
```suggestion
    size_t config_count = 0;
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,27 +3,377 @@
...
+            performance_mode);
+        return false;
+    }
+    SetNullTermPtrArray(power_configs, up_vote_power_configs_ptr_);
+
+    return true;
+  }
+
+  bool InitDownVotePowerConfigs(HtpPerformanceMode performance_mode) {
+    auto& power_configs = down_vote_power_configs_.emplace();
```
weilhuan-qti: I am not sure about the profit of using `std::optional` , `down_vote_power_configs_` is always constructed.

litert/vendors/qualcomm/core/backends/dsp_backend.cc
graham0824: Yes, you're right, the dsp_perf_infra_ is a raw pointer, we should use assignment instead of move.

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Removed draft code with std::optional change

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Removed the draft code of using optional, thanks for the reminder!

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Thanks for the comment! I think we don't need to check the return value of InitXXXConfigs() as we already check it before calling it, let me remove the check and align with dsp_backend.cc.

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Discussed offline, change to below code:

`template <typename T, std::size_t N>
void SetNullTermPtrArray(absl::Span<const T> src,
                         std::array<const T*, N>& dst) {
  for (std::size_t i = 0; i < src.size(); ++i) {
    dst[i] = &src[i];
  }
  dst[src.size()] = nullptr;
}`

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,36 +3,375 @@
...
+
+  void UpVote() {
+    if (htp_perf_infra_) {
+      htp_perf_infra_->perfInfra.setPowerConfig(
+          power_config_id_, up_vote_power_configs_ptr_.data());
+    }
+  }
+
+  void DownVote() {
+    if (htp_perf_infra_ && !down_vote_power_configs_ptr_.empty()) {
```
weilhuan-qti: If we have some reason to check whether `down_vote_power_configs_ptr_` is empty, please leave some comments to explain.

litert/vendors/qualcomm/core/utils/backend_utils.h
```diff
@@ -0,0 +1,39 @@
...
+  static constexpr uint32_t kRpcPollingTimeHighPower = 9999;
+};
+
+template <typename T, std::size_t N>
+void SetNullTermPtrArray(absl::Span<const T> src,
+                         std::array<const T*, N>& dst) {
+  for (std::size_t i = 0; i < src.size(); ++i) {
+    dst[i] = &src[i];
+  }
+  dst[src.size()] = nullptr;
```
weilhuan-qti: We should check the size of `dst` too. For such util function, we can write unit test to verify it.
```suggestion
  size_t min_size = std::min(src.size(), dst.size() - 1);
  for (std::size_t i = 0; i < min_size ; ++i) {
    dst[i] = &src[i];
  }
  dst[min_size] = nullptr;
```

litert/vendors/qualcomm/core/backends/htp_backend.cc
```diff
@@ -3,36 +3,375 @@
...
-#include <optional>
 #include <utility>
 #include <vector>
 
 #include "absl/types/span.h"  // from @com_google_absl
-#include "litert/vendors/qualcomm/core/backends/htp_perf_control.h"
 #include "litert/vendors/qualcomm/core/backends/qnn_backend.h"
 #include "litert/vendors/qualcomm/core/common.h"
 #include "litert/vendors/qualcomm/core/schema/soc_table.h"
+#include "litert/vendors/qualcomm/core/utils/backend_utils.h"
```
weilhuan-qti: Open for discussion, I prefer to move this header into litert/vendors/qualcomm/core/backend. What do you think?

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: This code is set aside for manual voting mode, and since we don't want to enable it right now, I'll go ahead and remove this check. Thanks for the reminder!

litert/vendors/qualcomm/core/utils/backend_utils.h
graham0824: You're correct! It makes sense to include this condition in the util function, even though the current use case hasn't required it yet.

Added 2 test cases in utils_test to verify it.

litert/vendors/qualcomm/core/backends/htp_backend.cc
graham0824: Since backend_utils.h is only utilized in the backend implementation, I agree that we can move it.

litert/vendors/qualcomm/core/utils/utils_test.cc
```diff
@@ -168,4 +169,26 @@ TEST(MiscTests, DISABLED_LoadIrBackendApiTest) {
...
+  std::vector<int> src = {1, 2, 3, 4, 5};
+  std::array<const int*, 4> dst;
+  SetNullTermPtrArray(absl::MakeConstSpan(src), dst);
+
+  EXPECT_EQ(*dst[0], 1);
+  EXPECT_EQ(*dst[1], 2);
+  EXPECT_EQ(*dst[2], 3);
+  EXPECT_EQ(dst[3], nullptr);
+}
+
```
weilhuan-qti: Nit: could we move these tests into `core/backends/backend_utils_test.cc`. Follow the convention of most test.s

graham0824: Sure

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,300 @@
...
+  test_device_hardware_device_info = {};
+  test_device_platform_info = {};
+}
+
+static Qnn_ErrorHandle_t MockSetPowerConfig(
+    uint32_t power_config_id,
+    const QnnHtpPerfInfrastructure_PowerConfig_t** power_configs) {
+  std::vector<QnnHtpPerfInfrastructure_PowerConfig_t> local_configs;
+  if (power_configs) {
+    for (int i = 0; power_configs[i] != nullptr; ++i) {
```
weilhuan-qti:
```suggestion
    for (size_t i = 0; power_configs[i] != nullptr; ++i) {
```

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+  real_device_get_infrastructure = nullptr;
+  test_captured_configs.clear();
+  test_infra = {};
+}
+
+static Qnn_ErrorHandle_t MockSetPowerConfig(
+    uint32_t power_config_id,
+    const QnnDspPerfInfrastructure_PowerConfig_t** power_configs) {
+  if (power_configs) {
+    for (int i = 0; power_configs[i] != nullptr; ++i) {
```
weilhuan-qti:
```suggestion
    for (size_t i = 0; power_configs[i] != nullptr; ++i) {
```

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+namespace {
+
+// Use global variables here as QNN C API requires static function pointers.
+static QnnDevice_GetInfrastructureFn_t real_device_get_infrastructure = nullptr;
+static QnnDspDevice_Infrastructure_t test_infra = {};
+static std::vector<QnnDspPerfInfrastructure_PowerConfig_t>
+    test_captured_configs;
+
+// Mock functions.
+static void ResetMockState() {
```
weilhuan-qti: Actually `ResetMockState` is not a mock functions.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+  test_captured_configs.clear();
+  test_infra = {};
+}
+
+static Qnn_ErrorHandle_t MockSetPowerConfig(
+    uint32_t power_config_id,
+    const QnnDspPerfInfrastructure_PowerConfig_t** power_configs) {
+  if (power_configs) {
+    for (int i = 0; power_configs[i] != nullptr; ++i) {
+      test_captured_configs.push_back(*power_configs[i]);
```
weilhuan-qti: We can also clear `test_captured_configs` in this function.
```suggestion
      test_captured_configs.emplace_back(*power_configs[i]);
```

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+
+  auto* real_dsp_infra =
+      static_cast<QnnDspDevice_Infrastructure_t*>(real_handle);
+  test_infra = *real_dsp_infra;
+  test_infra.setPowerConfig = MockSetPowerConfig;
+
+  // The QNN API defines this parameter as const*, but it is actually an output
+  // parameter.
+  auto* output_ptr = const_cast<QnnDevice_Infrastructure_t*>(infra);
+  *output_ptr = static_cast<QnnDevice_Infrastructure_t>(&test_infra);
```
weilhuan-qti: I think you misunderstand the types here, we can discuss offline and make the implementation simpler.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+  auto handle = CreateDLHandle(DspBackend::GetLibraryName());
+  if (!handle) GTEST_SKIP();
+
+  auto* real_qnn_api =
+      ResolveQnnApi(handle.get(), DspBackend::GetExpectedBackendVersion());
+  ASSERT_TRUE(real_qnn_api);
+
+  QNN_INTERFACE_VER_TYPE test_qnn_api = *real_qnn_api;
+  real_device_get_infrastructure = test_qnn_api.deviceGetInfrastructure;
+  test_qnn_api.deviceGetInfrastructure = MockDeviceGetInfrastructure;
```
weilhuan-qti: Maybe we can directly modify `real_qnn_api`.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+
+  Options options;
+  options.SetDspPerformanceMode(params.mode);
+  DspBackend backend(&test_qnn_api);
+
+  EXPECT_TRUE(backend.Init(options, std::nullopt));
+
+  const auto& configs = test_captured_configs;
```
weilhuan-qti: I think we can rename `test_captured_configs` into `captured_configs`, it is shorter but meaningful.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,300 @@
...
+      QNN_DEVICE_HARDWARE_DEVICE_INFO_VERSION_1;
+  test_device_hardware_device_info.v1.deviceId = 0;
+  test_device_hardware_device_info.v1.numCores = 1;
+  test_device_hardware_device_info.v1.cores = &test_device_core_info;
+  test_device_hardware_device_info.v1.deviceInfoExtension =
+      &test_htp_device_info_extension;
+
+  test_device_platform_info.version = QNN_DEVICE_PLATFORM_INFO_VERSION_1;
+  test_device_platform_info.v1.numHwDevices = 1;
+  test_device_platform_info.v1.hwDevices = &test_device_hardware_device_info;
```
weilhuan-qti: These can be done when initialize `test_device_platform_info` because they won't be changed later.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,300 @@
...
+
+static Qnn_ErrorHandle_t MockDeviceFree(void* handle) {
+  if (handle == &test_device_platform_info) {
+    return QNN_SUCCESS;
+  }
+  if (real_device_free) {
+    return real_device_free(handle);
+  }
+  return QNN_SUCCESS;
+}
```
weilhuan-qti: Is it possible to enter here in our test cases?

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+class DspBackendPerfParamTest : public testing::TestWithParam<DspPerfParams> {
+ protected:
+  // Google Test calls SetUp before each test case.
+  void SetUp() override { ResetMockState(); }
+};
+
+TEST_P(DspBackendPerfParamTest, InitWithPerfMode) {
+  const auto& params = GetParam();
+
+  auto handle = CreateDLHandle(DspBackend::GetLibraryName());
```
ChunHsuehLee: If I remember correctly, `CreateDLHandle` caused issue when integrating with Google side. So, tests in `qnn_backend_test.cc` are all disabled.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,164 @@
...
+
+  Options options;
+  options.SetDspPerformanceMode(params.mode);
+  DspBackend backend(&test_qnn_api);
+
+  EXPECT_TRUE(backend.Init(options, std::nullopt));
+
+  const auto& configs = test_captured_configs;
+  ASSERT_FALSE(configs.empty());
+  ASSERT_GE(configs.size(), 6);
```
ChunHsuehLee: Is there any case that the config size will be larger than 6?

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,300 @@
...
+  QnnHtpPerfInfrastructure_VoltageCorner_t voltage_corner;
+};
+
+class HtpBackendPerfBaseTest : public testing::TestWithParam<HtpPerfParams> {
+ public:
+  void SetUp() override { ResetMockState(); }
+};
+
+class HtpBackendRPCPollingPerfParamTest : public HtpBackendPerfBaseTest {};
+class HtpBackendRPCControlPerfParamTest : public HtpBackendPerfBaseTest {};
```
ChunHsuehLee:
```suggestion
```

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,300 @@
...
+
+TEST_P(HtpBackendRPCPollingPerfParamTest, InitWithPerfMode) {
+  const auto& params = GetParam();
+
+  auto handle = CreateDLHandle(HtpBackend::GetLibraryName());
```
ChunHsuehLee: Ditto.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,300 @@
...
+    Qnn_LogHandle_t logger, const QnnDevice_PlatformInfo_t** platformInfo) {
+  test_htp_device_info_extension.devType = QNN_HTP_DEVICE_TYPE_ON_CHIP;
+  test_htp_device_info_extension.onChipDevice.arch = QNN_HTP_DEVICE_ARCH_V69;
+  test_htp_device_info_extension.onChipDevice.socModel = 35;
+
+  test_device_hardware_device_info.version =
+      QNN_DEVICE_HARDWARE_DEVICE_INFO_VERSION_1;
+  test_device_hardware_device_info.v1.deviceId = 0;
+  test_device_hardware_device_info.v1.numCores = 1;
+  test_device_hardware_device_info.v1.cores = &test_device_core_info;
```
ChunHsuehLee: Is `test_device_core_info` always `{}`?

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: Changed to use emplace_back.

But we can't call `captured_configs.clear();` in `MockSetPowerConfig`, since we will call `setPowerConfig()` twice in >= v69 case (One for RPC polling, another one for RPC control latency). Keep the clear() in `ResetMockState`.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: Sure, changed to below codes:

`  real_device_get_infrastructure = real_qnn_api->deviceGetInfrastructure;
  real_qnn_api->deviceGetInfrastructure = MockDeviceGetInfrastructure;`

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
graham0824: Yes, you're right! Changed to assign values in the initialization phase.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
graham0824: As we change to assign values for test_device_platform_info in the initialization phase, we could remove these codes and return ERROR here.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: Discussed offline, label with DISABLED and only test it internally.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: Thanks for the comment! Changed to use ASSERT_EQ here

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
graham0824: Actually, we have 2 kinds of test cases, one is for rpc polling with 2 configs, the other one is for rpc control with 1 config.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
graham0824: Thanks for the comment! Changed to initialize it with valid values.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: Discussed offline, change to use infra directly.

litert/vendors/qualcomm/core/backends/ir_backend_test.cc
```diff
@@ -50,29 +33,18 @@ class QnnBackendTest : public testing::TestWithParam<BackendType> {
 };
 
 INSTANTIATE_TEST_SUITE_P(QnnBackendTest, QnnBackendTest,
-                         testing::Values(BackendType::kHtpBackend,
-                                         BackendType::kIrBackend,
-                                         BackendType::kDspBackend));
+                         testing::Values(BackendType::kIrBackend));
```
weilhuan-qti: Maybe we don't need parameterized test if there is only ir backend. And we don't need switch below.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,200 @@
...
+
+  EXPECT_EQ(configs[0].config,
+            QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_SLEEP_LATENCY);
+  EXPECT_EQ(configs[0].sleepLatencyConfig, params.expected_sleep);
+  EXPECT_EQ(configs[1].config,
+            QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER);
+  EXPECT_EQ(configs[1].dcvsVoltageCornerMinConfig, params.expected_min_voltage);
+  EXPECT_EQ(configs[2].config,
+            QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER);
+  EXPECT_EQ(configs[2].dcvsVoltageCornerMinConfig,
```
weilhuan-qti:
```suggestion
  EXPECT_EQ(configs[2].dcvsVoltageCornerTargetConfig,
```

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,200 @@
...
+            params.expected_target_voltage);
+  EXPECT_EQ(configs[3].config,
+            QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER);
+  EXPECT_EQ(configs[3].dcvsVoltageCornerMinConfig, params.expected_max_voltage);
```
weilhuan-qti:
```suggestion
  EXPECT_EQ(configs[3].dcvsVoltageCornerMaxConfig, params.expected_max_voltage);
```

litert/vendors/qualcomm/core/backends/dsp_backend.cc
```diff
@@ -3,32 +3,219 @@
...
+    power_configs[kDcvsVcornerMin].dcvsVoltageCornerMinConfig = voltage_min;
+
+    power_configs[kDcvsVcornerTarget].config =
+        QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER;
+    power_configs[kDcvsVcornerTarget].busVoltageCornerTargetConfig =
+        voltage_target;
+
+    power_configs[kDcvsVcornerMax].config =
+        QNN_DSP_PERF_INFRASTRUCTURE_POWER_CONFIGOPTION_DCVS_VOLTAGE_CORNER;
+    power_configs[kDcvsVcornerMax].busVoltageCornerMaxConfig = voltage_max;
```
weilhuan-qti: Seems we use the wrong member here. It is weird that we set partial setting for dcvs and bus.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,200 @@
...
+  DLHandle handle_;
+  std::unique_ptr<DspBackend> backend_;
+};
+
+TEST_F(DspBackendTest, DISABLED_InitializeWithLogLevelOffTest) {
+  Options options;
+  options.SetLogLevel(LogLevel::kOff);
+
+  ASSERT_TRUE(backend_->Init(options, std::nullopt));
+  ASSERT_FALSE(backend_->GetDeviceHandle());
```
weilhuan-qti, line 178: Can we make the style consistent with htp_backend_test.cc
```suggestion
  ASSERT_EQ(backend_->GetDeviceHandle(), nullptr);
```

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,324 @@
...
+Qnn_ErrorHandle_t MockSetPowerConfig(
+    uint32_t power_config_id,
+    const QnnHtpPerfInfrastructure_PowerConfig_t** power_configs) {
+  std::vector<QnnHtpPerfInfrastructure_PowerConfig_t> local_configs;
+  if (power_configs) {
+    for (size_t i = 0; power_configs[i] != nullptr; ++i) {
+      local_configs.emplace_back(*power_configs[i]);
+    }
+  }
+  captured_configs.emplace_back(local_configs);
```
weilhuan-qti:
```suggestion
  captured_configs.emplace_back(std::move(local_configs));
```

litert/vendors/qualcomm/core/backends/ir_backend_test.cc
graham0824: Thanks for the reminder! Removed this for ir_backend_test

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: Nice catch!

litert/vendors/qualcomm/core/backends/dsp_backend.cc
graham0824: Discussed offline, change to set dvcs config

litert/vendors/qualcomm/core/backends/backend_utils_test.cc
```diff
@@ -0,0 +1,59 @@
...
+namespace qnn {
+namespace {
+TEST(BackendUtilsTest, SetNullTermPtrArrayTest) {
+  std::vector<int> src = {1, 2, 3};
+  std::array<const int*, 5> dst;
+  SetNullTermPtrArray(absl::MakeConstSpan(src), dst);
+
+  EXPECT_EQ(*dst[0], 1);
+  EXPECT_EQ(*dst[1], 2);
+  EXPECT_EQ(*dst[2], 3);
```
KelvinYang0320:
```suggestion
  std::array<const int*, 5> dst{};
  SetNullTermPtrArray(absl::MakeConstSpan(src), dst);

  ASSERT_EQ(dst[0]);
  EXPECT_EQ(*dst[0], 1);
  ASSERT_EQ(dst[1]);
  EXPECT_EQ(*dst[1], 2);
  ASSERT_EQ(dst[2]);
  EXPECT_EQ(*dst[2], 3);
```

litert/vendors/qualcomm/core/backends/backend_utils_test.cc
```diff
@@ -0,0 +1,59 @@
...
+  EXPECT_EQ(dst[3], nullptr);
+}
+
+TEST(BackendUtilsTest, FindSocInfoTest) {
+  // Test finding a valid SoC (SDM865)
```
KelvinYang0320: Nit:
```suggestion
  // Test finding a valid SoC (SDM865).
```

litert/vendors/qualcomm/core/backends/backend_utils_test.cc
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Qualcomm Innovation Center, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "litert/vendors/qualcomm/core/backends/backend_utils.h"
+
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <optional>
+#include <vector>
```
KelvinYang0320, line 12: Could you include #include "absl/types/span.h"  and #include <arracy>?

litert/vendors/qualcomm/core/backends/backend_utils_test.cc
```diff
@@ -0,0 +1,59 @@
...
+
+namespace qnn {
+namespace {
+TEST(BackendUtilsTest, SetNullTermPtrArrayTest) {
+  std::vector<int> src = {1, 2, 3};
+  std::array<const int*, 5> dst;
+  SetNullTermPtrArray(absl::MakeConstSpan(src), dst);
+
+  EXPECT_EQ(*dst[0], 1);
```
KelvinYang0320: If `SetNullTermPtrArray` does not correctly assign the pointer to `dst`, all pointers in `dst` will remain uninitialized.

Could you update the code to use `std::array<const int*, 5> dst{};` and ensure that `dsp[i]` is not a `nullptr` before dereferencing it?

litert/vendors/qualcomm/core/backends/backend_utils_test.cc
```diff
@@ -0,0 +1,59 @@
...
+  EXPECT_EQ(*dst[1], 2);
+  EXPECT_EQ(*dst[2], 3);
+  EXPECT_EQ(dst[3], nullptr);
+}
+
+TEST(BackendUtilsTest, SetNullTermPtrArrayTestOverflow) {
+  std::vector<int> src = {1, 2, 3, 4, 5};
+  std::array<const int*, 4> dst;
+  SetNullTermPtrArray(absl::MakeConstSpan(src), dst);
```
KelvinYang0320, line 33: ditto

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
```diff
@@ -0,0 +1,200 @@
...
+};
+
+class DspBackendPerfParamTest : public testing::TestWithParam<DspPerfParams> {
+ public:
+  void SetUp() override {
+    handle_ = CreateDLHandle(DspBackend::GetLibraryName());
+    if (!handle_) GTEST_SKIP();
+
+    real_qnn_api_ = const_cast<QNN_INTERFACE_VER_TYPE*>(
+        ResolveQnnApi(handle_.get(), DspBackend::GetExpectedBackendVersion()));
```
KelvinYang0320: Would it be more secure to copy the entire qnn interface instead of using `const_cast`?

For example, change to `QNN_INTERFACE_VER_TYPE qnn_api_copy_{};`
and then
```c++
const QNN_INTERFACE_VER_TYPE* qnn_api = ResolveQnnApi(...);
qnn_api_copy_ = *qnn_api;
```

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,327 @@
...
+QnnDevice_HardwareDeviceInfo_t test_device_hardware_device_info = {
+    .version = QNN_DEVICE_HARDWARE_DEVICE_INFO_VERSION_1,
+    .v1 = {.deviceId = 0,
+           .numCores = 1,
+           .cores = &test_device_core_info,
+           .deviceInfoExtension = &test_htp_device_info_extension}};
+
+QnnDevice_PlatformInfo_t test_device_platform_info = {
+    .version = QNN_DEVICE_PLATFORM_INFO_VERSION_1,
+    .v1 = {.numHwDevices = 1, .hwDevices = &test_device_hardware_device_info}};
```
KelvinYang0320, line 47: Would it be possible to declare these variables as `constexpr`, or at least mark them as `const`?
And we can change to naming like `kTestDevicePlatformInfo`.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
```diff
@@ -0,0 +1,327 @@
...
+  void SetUp() override {
+    // Clean the previous captured configs.
+    captured_configs.clear();
+
+    handle_ = CreateDLHandle(HtpBackend::GetLibraryName());
+    if (!handle_) GTEST_SKIP();
+
+    real_qnn_api_ = const_cast<QNN_INTERFACE_VER_TYPE*>(
+        ResolveQnnApi(handle_.get(), HtpBackend::GetExpectedBackendVersion()));
+    ASSERT_TRUE(real_qnn_api_);
```
KelvinYang0320: Similar suggestion as dsp_backend_test.

litert/vendors/qualcomm/core/backends/backend_utils_test.cc
graham0824: Nice catch! We should add nullptr check before dereferencing it.

litert/vendors/qualcomm/core/backends/dsp_backend_test.cc
graham0824: By changing to modify the copied api, it will be safer to not impact the original API and we could simplify the implmentation in TearDown(), updated in the latest commit.

litert/vendors/qualcomm/core/backends/htp_backend_test.cc
graham0824: Discussed offline, change line 45 to constexpr

litert/vendors/qualcomm/core/backends/ir_backend_test.cc
```diff
@@ -0,0 +1,53 @@
...
+#include <gtest/gtest.h>
+#include "litert/vendors/qualcomm/core/backends/ir_backend.h"
+#include "litert/vendors/qualcomm/core/common.h"
+#include "litert/vendors/qualcomm/core/schema/soc_table.h"
+#include "litert/vendors/qualcomm/core/utils/miscs.h"
+
+namespace qnn {
+namespace {
+
+class QnnBackendTest : public testing::Test {
```
weilhuan-qti:
```suggestion
class IrBackendTest : public testing::Test {
```

PR #5261, opened 2026-01-13
Do not check if SDK path start with "/" on Windows.

taste1981:
local_path_env will not specify a path starting with "/" on Windows, so never check for that.


taste1981:
@reillyeon could you help to review this PR? Thanks!

reillyeon:
> @reillyeon could you help to review this PR? Thanks!

I do not have commit privileges to this project.

PR #5564, opened 2026-02-01
Update ov

JeevakaPrabu:


fujunwei:
LGTM.

litert/vendors/intel_openvino/dispatch/invocation_context.cc
```diff
@@ -54,10 +54,7 @@ LiteRtDispatchInvocationContextT::Create(
     return litert::Error(kLiteRtStatusErrorRuntimeFailure,
                          "Failed to get OpenVINO core from device context");
   }
-  ov::AnyMap configs_map;
-  configs_map["NPU_TURBO"] = "YES";
-  ov::CompiledModel compiled_model =
-      core->import_model(model_stream, "NPU", configs_map);
+  ov::CompiledModel compiled_model = core->import_model(model_stream, "NPU");
```
MatthiasKreileder, line 57: Thanks for the fix.

> it causes power regression

You explained it in the commit message, but for my eduction, could you elaborate a bit more? As long as we run with https://github.com/google-ai-edge/LiteRT/blob/1cec812cccce771380c12b040bdf89d973828232/litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc#L367 we are configuring the NPU correctly to minimize latency?

third_party/intel_openvino/openvino.bzl
```diff
@@ -28,27 +28,27 @@ def openvino_configure():
...
                 "host_os": "linux",
                 "file_extension": "tgz",
                 "symlink_mapping": {
-                    "openvino": "openvino_toolkit_ubuntu22_2025.4.0.dev20251113_x86_64",
+                    "openvino": "openvino_toolkit_ubuntu22_2026.0.0.dev20260127_x86_64",
                 },
             },
             {
-                "url": "https://storage.openvinotoolkit.org/repositories/openvino/packages/pre-release/2025.4.0rc2/openvino_toolkit_android_2025.4.0.dev20251113_x86_64.tgz",
+                "url": "https://storage.openvinotoolkit.org/repositories/openvino/packages/pre-release/2026.0.0rc1/openvino_toolkit_android_2026.0.0.dev20260127_x86_64.tgz",
```
MatthiasKreileder: Just in case we need to repro internally, do you know (can be approximately) which openvino commit this corresponds to?

This still doesn't contain `libopenvino_intel_npu_plugin.so`, right? Do you know the rough timeline for this?

litert/vendors/intel_openvino/dispatch/invocation_context.cc
akodanka: The performance hint LATENCY indicates the inference engine(software) to prioritize(minimize) single inference time over the throughput.
But the NPU_TURBO is a hardware configuration that changes several power management controls to obtain better perf of NPU.

third_party/intel_openvino/openvino.bzl
JeevakaPrabu: @MatthiasKreileder Incase of nightly build, commit hash is suffixed to the package folder name. for eg:  3ef8425ae12 is the commit hash corresponding to the build https://storage.openvinotoolkit.org/repositories/openvino/packages/nightly/2026.1.0-20996-3ef8425ae12/openvino_toolkit_ubuntu22_2026.1.0.dev20260131_x86_64.tgz .

npu plugin library is part of this nightly build

PR #5975, opened 2026-02-20
[intel_openvino] Add support NPU and platform version names.

anikulk:


anikulk:
Hi @MatthiasKreileder , Would soc model always be passed with LiteRtStatus LiteRtCompilerPluginCompile(
    LiteRtCompilerPlugin compiler_plugin, const char* soc_model,
    LiteRtModel partitions, LiteRtCompiledResult* compiled_result) . It appears benchmark_model doesnt accept the soc model flag. By default we allow the model to attempt compile in the benchmark_model case where soc name is not passed to the plugin.  The current change introduces the different NPU variants based on the version number per platform. We are not using it to compile the model but just to suggest that this model is supported.

MatthiasKreileder:
Thanks for the pull request!

> Would soc model always be passed 

It depends on the compilation mode:

**AOT compilation**
- User invokes apply_plugin_main (most likely indirectly via a Pythong API) with a target SoC
- This gets forwarded to the partition and compile functions of the vendor. E.g. https://github.com/google-ai-edge/LiteRT/blob/1b4e8b5eb8945cd062cea72208ba6f911f27e662/litert/tools/apply_plugin.cc#L326 
- Expected vendor implementation: Only partition(claim) / compile OPs if provided target SoC is supported

**JIT compilation**
- LiteRT runtime calls partition and compile functions with soc_model as nullptr https://github.com/google-ai-edge/LiteRT/blob/1b4e8b5eb8945cd062cea72208ba6f911f27e662/litert/runtime/compiled_model.cc#L602 
- Expected vendor implementation: Detect device's chipset and partition (claim) / compile OPs if the current device’s SoC is supported

Same for LiteRtCompilerPluginCheckCompilerCompatibility (for AOT soc_model is provided, for JIT the plugin needs to detect it itself)

MatthiasKreileder (changes requested):
Please take a look at https://github.com/google-ai-edge/LiteRT/pull/5975#issuecomment-3954849787

Please also consider adding a csv file like we have for other vendors https://github.com/google-ai-edge/LiteRT/blob/main/litert/vendors/mediatek/supported_soc.csv that calls out soc_model to (openvino) SDK version recommendation / compatibility

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
MatthiasKreileder, line 327: If `soc_model` is provided during ahead-of-time (AOT) compilation you should check if that `soc_model` is supported. If it is not supported, you should not claim any nodes as supported.

During just-in-time compilation (JIT) `soc_model` will be a nullptr and you determine if the chipset the executable is currently running on is supported. Presumably the underlying OpenVino SDK has such logic anyways? If detection of the current chipset here within the openvino_compiler_plugin.cc‎ is complicated/unreasonable effort (i.e. it's not easy to detect here) it's also _ok_ for now to claims nodes purely based on the list of supported OPs and then fail/pass in the compilation stage. Please leave a TODO in the code in this case.

You can check how this is done for MTK https://github.com/google-ai-edge/LiteRT/blob/9c00e9160448de44983edea4bac2b8b8cef6053a/litert/vendors/mediatek/compiler/compiler_plugin.cc#L363, essentially treating soc_model as an optional argument.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -341,6 +361,13 @@ LiteRtStatus LiteRtCompilerPluginCompile(
 
       // Configure device type
       auto device_type = intel_opts.GetDeviceType();
+      if (device_type == kLiteRtIntelOpenVinoDeviceTypeNPU) {
+        if (!IsSocModelSupported(soc_model)) {
```
MatthiasKreileder: `soc_model` will be a `nullptr` for just-in-time compilation. In this case we should not return with an error but have your code determine if the current/underlying chipset is supported.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -400,6 +427,12 @@ LiteRtStatus LiteRtCompilerPluginCompile(
           break;
       }
     } else {
+      if (!IsSocModelSupported(soc_model)) {
+        LITERT_LOG(LITERT_INFO,
+                   "Unsupported Intel NPU SoC model: %s Compile will proceed "
```
MatthiasKreileder: Similar to above, wouldn't this now always print "Unsupported" for the just-in-time compilation case where `soc_model` is a `nullptr`?

litert/vendors/intel_openvino/supported_soc.csv
```diff
@@ -0,0 +1,4 @@
+# manufacturer,NPU version, Platform name, Recommend OpenVINO Version,Compatible OpenVINO Version
+Intel, NPU27xx, Intel NPU for Intel Core Ultra processors codename Meteor Lake, 2026.0, "2026.0, 2025.4"
```
MatthiasKreileder: `NPU27xx` is a wildcard, right? Would it be possible to enumerate the exact chipset names that are supported instead? Presumably that's quite a bit more to type/add in this file, but it would be more specific and help developers what specific `soc` names they can specify for AOT compilation.

anikulk: Hi @MatthiasKreileder , 

Latest changes dont use wildcard. In supported_soc.csv we provide the platform names, which developers can access and the corresponding soc_name i.e PTL or LNL. 

Currently we keep the SOC names geneic like Series 2 and Series 3. In further changes we will provide the specific SKU names for example: https://www.intel.com/content/www/us/en/products/sku/245722/intel-core-ultra-7-processor-355-12m-cache-up-to-4-70-ghz/specifications.html 
  
  Also Note that generally the delegate/openvino and underlying NPU stack should work on any Series3/2 devices but we will validate as well.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
anikulk: Fixed this logic.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
anikulk: In the JIT flow not specifying NPU_PLATFORM will compile for the underlying platform.

litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -19,6 +19,21 @@ package(
     default_visibility = ["//litert:__subpackages__"],
 )
 
+litert_lib(
+    name = "openvino_compiler_plugin_internal",
+    srcs = ["openvino_compiler_plugin_internal.cc"],
+    hdrs = ["openvino_compiler_plugin_internal.h"],
```
MatthiasKreileder: I suppose openvino_compiler_plugin_internal.h and openvino_compiler_plugin_internal.cc need to be added to this pull request, right?

anikulk: Yes apologies i missed adding it earlier - done now.

litert/vendors/intel_openvino/supported_soc.csv
```diff
@@ -0,0 +1,3 @@
+# manufacturer,SOC Name, NPU version, Platform name, Recommend OpenVINO Version,Compatible OpenVINO Version
+Intel, LNL, NPU4000, Intel NPU for Intel Core Ultra processors (Series 2) codename Lunar Lake, 2026.0, "2026.0, 2025.4"
+Intel, PTL, NPU5010, Intel NPU for Intel Core Ultra processors (Series 3) codename Panther Lake, 2026.0, "2026.0, 2025.4"
```
JeevakaPrabu: As all 2026 versions will be compatible with LNL, PTL, do we keep it as 2026.x or we keep updating it on every OV release?

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -27,13 +30,6 @@
 #include <utility>
 #include <vector>
 
-#include "openvino/core/any.hpp"
-#include "openvino/core/except.hpp"
```
akodanka, line 31: Can you check if the exceptions in the compiler plugin are getting caught even after this change.
We had faced issues with cros-library exception handling :
https://github.com/google-ai-edge/LiteRT/pull/6015/changes/1ce6ea0490aee72bcfd5754327e71d94faff902d

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin_internal.cc
```diff
@@ -0,0 +1,66 @@
...
+
+#include "litert/c/internal/litert_logging.h"
+#include "litert/c/litert_common.h"
+#include "litert/cc/litert_macros.h"
+#include "openvino/core/any.hpp"
+
+namespace litert::openvino {
+
+// Map from SoC model codename to the NPU platform string used by OpenVINO.
+const std::map<const char*, const char*> kSocModelConfigs = {
```
akodanka: Can you consider using unordered_map ?
It can help accessing elements in the below functions too.

litert/vendors/intel_openvino/supported_soc.csv
anikulk: If this is garanteed its better to keep as 2026.x. Ill make the change.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
anikulk: Hi Anoob, that was an error in rebase, I added the ov except.hpp include. Additionally removed unused headers.

PR #6232, opened 2026-03-09
Fix Windows MSVC build for //litert/tools:benchmark_model

taste1981:
Address multiple issues preventing the Bazel build from succeeding on Windows with MSVC (cl.exe) and C++20:

1. Move GCC/Clang-only flags (-std=gnu++17, -fpermissive) from unconditional 'build' to platform-specific 'build:linux'/'build:macos' sections in .bazelrc so they don't get passed to cl.exe.

2. Add 'build:windows --cpu=x64_windows' so that XLA/TSL config_settings (which match on cpu=x64_windows) resolve correctly. Without this, if_not_windows() selects fall through to the default branch and emit GCC flags like -Wno-sign-compare to MSVC.

3. Patch XLA's tsl.bzl (via PATCH.tf_xla_tsl_win_copts) to fix get_win_copts(is_msvc=False) -> is_msvc=True, so TSL emits MSVC-style compiler flags on Windows.

4. Add per_file_copt to compile protobuf sources with /std:c++17 on Windows. Protobuf v6.31.1 uses incomplete (forward-declared) types as flat_hash_map value types, which fails under MSVC C++20 due to P0848R3 requiring complete types for conditionally-trivial special members.

5. Patch protobuf's port.h (via PATCH.protobuf_port_msvc_compat) to force GlobalEmptyString to use GlobalEmptyStringDynamicInit on MSVC. This prevents an ABI mismatch where protobuf .cc files compiled at C++17 define a different mangled symbol than .pb.cc files compiled at C++20.

6. Extend tensorflow_source_repo rule to support 'patches' and 'protobuf_patches' attributes, allowing patches to be applied to the TensorFlow source and appended to XLA's protobuf.patch respectively.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -52,10 +52,14 @@ build:clang_local --repo_env USE_HERMETIC_CC_TOOLCHAIN=1
 
 build --repo_env=USE_PYWRAP_RULES=True
 build --copt=-DGRPC_BAZEL_BUILD
-build --cxxopt=-std=gnu++17
-build --cxxopt=-fpermissive
 build --host_copt=-DGRPC_BAZEL_BUILD
-build --host_cxxopt=-fpermissive
+# GCC/Clang-specific flags: only apply on non-Windows platforms.
+# Each platform already specifies its own C++ standard (see below).
+build:linux --cxxopt=-std=gnu++17
+build:linux --cxxopt=-fpermissive
+build:linux --host_cxxopt=-fpermissive
+build:macos --cxxopt=-fpermissive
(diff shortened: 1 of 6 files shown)


MatthiasKreileder:
Closing with license files added https://github.com/google-ai-edge/LiteRT/commit/074e46c91781eec161ca17c4a6ce7d5f3ab70e8a#diff-91a112573b079cf4f657bd518ccb3ff6ff2be467dfa506a14f24cb7bd8a302c3 (git history preserved that you authored this change)

PR #6330, opened 2026-03-12
[Windows] Disable OpenCL support on Windows builds

enlenama:
This modification explicitly disables OpenCL support in Windows builds to
maintain alignment and dependency compatibility with LiteRT-LM.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -282,6 +282,9 @@ build:windows --host_copt=/W0
 # GCC/Clang flags (e.g. -Wno-sign-compare) to MSVC.
 build:windows --cpu=x64_windows
 
+# Exclude OpenCL from Windows builds.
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -282,6 +282,9 @@ build:windows --host_copt=/W0
 # GCC/Clang flags (e.g. -Wno-sign-compare) to MSVC.
 build:windows --cpu=x64_windows
 
+# Exclude OpenCL from Windows builds.
+build:windows --copt=/DLITERT_DISABLE_OPENCL_SUPPORT=1
```
protobird-git, line 286: It should be disabled by default. https://github.com/google-ai-edge/LiteRT/blob/main/litert/c/litert_common.h#L146

YingyingMaIntel: Thank you for the feedback! I see it's already been handled on Windows, my modification is out of date, will close this PR.

PR #6445, opened 2026-03-17
eTPU for 3p: Use linker-visible eTPU soname and add FD load fallback.

bpkeene:
Switch Android Southbound loading to `libedgetpu_litert.so` and add an mmap-based fallback when `thrLoadSqContainerFdWithOffset` is unavailable so eTPU dispatch works on older runtimes.

### DIFF
--- a/litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
+++ b/litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
@@ -18,6 +18,8 @@
 
 #include <cstddef>
 #include <optional>
+#include <sys/mman.h>
+#include <unistd.h>
 #include <utility>
 
 #if __ANDROID__
@@ -40,6 +42,7 @@
 #include "litert/vendors/google_tensor/dispatch/dispatch_api_config.h"
 #include "litert/vendors/google_tensor/dispatch/dispatch_api_macros.h"
 #include "litert/vendors/google_tensor/dispatch/sb_api.h"
+#include "litert/vendors/google_tensor/dispatch/sb_api_features.h"
 
 namespace gt = litert::google_tensor;
 
@@ -86,6 +89,14 @@ LiteRtStatus LiteRtDispatchDeviceContextT::Destroy() {
     LITERT_LOG(LITERT_ERROR,
                "Cannot destroy device context with %zu graphs registered",
                registered_graphs_.size());
+    return kLiteRtStatusErrorRuntimeFailure;
+  }
+
+  if (!mmap_regions_.empty()) {
+    LITERT_LOG(LITERT_ERROR,
+               "Cannot destroy device context with %zu mmap'd executable(s) "
+               "still loaded",
+               mmap_regions_.size());
     return kLiteRtStatusErrorRuntimeFailure;
   }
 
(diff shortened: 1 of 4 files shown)


jl45G (commented):
change lgtm, please rebase the code, thank you1

jl45G (approved):

bpkeene:
merge?

jl45G:
Some internal only testings are failing, let me fix those

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
```diff
@@ -196,13 +196,38 @@ LiteRtStatus LiteRtDispatchDeviceContextT::LoadExecutable(
...
 
   if (bytecode_buffer.fd >= 0) {
-    GT_LOG_RETURN_IF_SB_ERROR(
-        thrLoadSqContainerFdWithOffset(thr_context_, thr_type,
-                                       bytecode_buffer.fd, bytecode_buffer.size,
-                                       bytecode_buffer.offset,
-                                       /*lazy_loading=*/false, &exec_handle),
-        "Failed to load SQ container from fd %d with size %zu and offset %zu",
-        bytecode_buffer.fd, bytecode_buffer.size, bytecode_buffer.offset);
+    ThrStatus sq_status = thrLoadSqContainerFdWithOffset(
```
jl45G: sb_api_features.h is how we manage backwards compatibility with the SouthBound shared object.

Can you please use this mechanism? 

thrLoadSqContainerFdWithOffset can fail for a variety of reasons other than the symbol being absent from the SouthBound shared object in the vendor image.

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
```diff
@@ -196,13 +196,38 @@ LiteRtStatus LiteRtDispatchDeviceContextT::LoadExecutable(
...
+        thr_context_, thr_type, bytecode_buffer.fd, bytecode_buffer.size,
+        bytecode_buffer.offset, /*lazy_loading=*/false, &exec_handle);
+    if (sq_status != kThrStatusSuccess) {
```
jl45G: If offset == 0, we can fallback to thrLoadSqContainerFd, which is present in old SouthBound shared objects.

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
```diff
@@ -196,13 +196,38 @@ LiteRtStatus LiteRtDispatchDeviceContextT::LoadExecutable(
...
+      // thrLoadSqContainerFdWithOffset: mmap the bytecode region from the
+      // flatbuffer FD and load via the buffer-based API.
+      LITERT_LOG(LITERT_INFO,
+                 "thrLoadSqContainerFdWithOffset unavailable or failed, "
+                 "falling back to mmap + thrLoadSqContainer");
+      size_t page_size = static_cast<size_t>(sysconf(_SC_PAGESIZE));
+      size_t aligned_offset = bytecode_buffer.offset & ~(page_size - 1);
+      size_t offset_delta = bytecode_buffer.offset - aligned_offset;
+      size_t map_length = bytecode_buffer.size + offset_delta;
+      void* mapped = mmap(nullptr, map_length, PROT_READ, MAP_PRIVATE,
```
jl45G: This should only be needed if offset > 0. What is the case that has a non-zero offset? It would be nice to avoid all of this boilerplate. We also have a class in internal codebase that is designed to do exactly this, and is thoroughly tested. If needed, we could look at bringing that class in to our Dispatch API implementation here.

bpkeene: this is the offset of the DISPATCH_OP containing the eTPU bytecode within the AOT compiled LiteRT model targeting G3/G4/G5 - our file in this use case is the .tflite flatbuffer, not raw eTPU program.

I do not know what / any guarantees the compiler or ai-edge-litert via AOT sdk has for placement of the dispatch op within the flatbuffers regarding the offset vs page alignment

jl45G: Thank you for the response, I have invited the google tensor team to directly review this PR

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
```diff
@@ -196,13 +196,38 @@ LiteRtStatus LiteRtDispatchDeviceContextT::LoadExecutable(
...
+        LITERT_LOG(LITERT_ERROR,
+                   "Failed to mmap SQ bytecode (fd=%d, size=%zu, offset=%zu)",
+                   bytecode_buffer.fd, bytecode_buffer.size,
+                   bytecode_buffer.offset);
+        return kLiteRtStatusErrorRuntimeFailure;
+      }
+      const void* bytecode_ptr =
+          static_cast<const std::byte*>(mapped) + offset_delta;
+      GT_LOG_RETURN_IF_SB_ERROR(
+          thrLoadSqContainer(thr_context_, thr_type, bytecode_ptr,
```
tywerbicki-cmd: If this fails, it seems like we're leaking the memory map created above.

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.h
```diff
@@ -66,22 +69,26 @@ class LiteRtDispatchDeviceContextT {
...
+    size_t length;
+  };
+
   ThrContext* absl_nonnull thr_context_;
 #if LITERT_HAS_DARWINN_OPTIONS_SUPPORT
   std::optional<litert::LiteRtDarwinnRuntimeOptionsT> darwinn_options_;
 #endif  // LITERT_HAS_DARWINN_OPTIONS_SUPPORT
   // A device context cannot be destroyed with any registered graphs.
   absl::flat_hash_set<LiteRtDispatchGraph> registered_graphs_;
+  std::vector<MmapRegion> mmap_regions_;
```
tywerbicki-cmd, line 98: Can you elaborate on the choice of data structure?

Are you avoiding an associative container like `absl::flat_hash_set` because you expect the number of entries to be very small?

litert/vendors/google_tensor/dispatch/sb_api_late_binding.cc
```diff
@@ -185,7 +185,10 @@ T GetValForSymUnavailable() {
 
 constexpr char kDefaultSouthBoundLibPath[] =
 #ifdef __ANDROID__
-    "/vendor/lib64/libedgetpu_litert.so";
+    // Use bare soname instead of absolute vendor path so that Android's linker
+    // namespace resolves the library via public.libraries.txt. The absolute
+    // path is blocked for untrusted_app contexts on Android 12+.
+    "libedgetpu_litert.so";
```
tywerbicki-cmd, line 191: Appreciate the fix!

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
```diff
@@ -87,6 +81,12 @@ LiteRtStatus LiteRtDispatchDeviceContextT::Destroy() {
     return kLiteRtStatusErrorRuntimeFailure;
   }
 
+  // Release any remaining mmap'd bytecode regions.
+  for (auto& region : mmap_regions_) {
+    munmap(region.addr, region.length);
+  }
+  mmap_regions_.clear();
```
tywerbicki-cmd: Should this ever be non-empty here?

tywerbicki-cmd (changes requested):
All of @jl45G s feedback is originally from me.

I think most of it can be handled internally as "fix-forward" changes. However, I think some of it should be addressed prior to merge. In particular, using `thrLoadContainerFd` if `fd >= 0` but `offset == 0`.

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.h
bpkeene: No particular reason for vector vs hash set, but we would indeed expect the number of entries to be small (in our case, 1 always)

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
bpkeene: will revise and amend the PR with the above feedback as well

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.cc
bpkeene: Good question, will clarify & document relationship (must strictly live longer [before, after]) to registered_graphs

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.h
bpkeene: keeping vector in light of that.  

I don't know if eTPU driver has max program queue, or if LiteRT AOT sdks have a maximum number of dispatch ops permitted in a graph - in general though, we pre-process such that there is only 1 dispatch_op in a given .tflite graph.

bpkeene:
addressed the comments, please let me know if there are further requested changes

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.h
```diff
@@ -84,13 +87,24 @@ class LiteRtDispatchDeviceContextT {
   // Consumers of this class must use `Destroy` to delete the instance.
   ~LiteRtDispatchDeviceContextT() = default;
 
+  struct MmapRegion {
```
tywerbicki-cmd: Nit: google style dictates type definitions precede any function decl/defs. As such, can this be moved above the constructor definition?

litert/vendors/google_tensor/dispatch/litert_dispatch_device_context.h
tywerbicki-cmd: Thanks for the clarification!

tywerbicki-cmd (approved):
Nice PR. Thank you for the contribution, Brian!

PR #6461, opened 2026-03-18
Support exynos dispatch

Jiseong-oh:
Support dispatch of Litert for Exynos Delegate

- Support dispatch module of Exynos Delegate
- Add samsung options
- Support run_model

### DIFF
--- a/litert/c/windows_exported_symbols.def
+++ b/litert/c/windows_exported_symbols.def
@@ -208,3 +208,6 @@ EXPORTS
   LiteRtUnwrapDelegate
   LiteRtWaitEvent
   LiteRtWrapDelegate
+  LiteRtSamsungOptionsCreate
+  LiteRtSamsungOptionsGetIdentifier
+  LiteRtSamsungOptionsGet


litert/tools/run_model.cc
```diff
@@ -50,6 +51,7 @@
 #include "litert/tools/flags/vendors/intel_openvino_flags.h"  // IWYU pragma: keep
 #include "litert/tools/flags/vendors/mediatek_flags.h"  // IWYU pragma: keep
 #include "litert/tools/flags/vendors/qualcomm_flags.h"  // IWYU pragma: keep
+#include "litert/tools/flags/vendors/samsung_flags.h"  // IWYU pragma: keep
```
jl45G, line 54: could you also update the cmake

litert/vendors/samsung/dispatch/litert_dispatch_invocation_context.cc
```diff
@@ -0,0 +1,251 @@
...
+          model_id, &tmp_buf_ptr, &buffer_info) != ENN_RET_SUCCESS) {
+    if (enn_manager->Api().EnnCloseModel(model_id) != ENN_RET_SUCCESS) {
+      return litert::Error(kLiteRtStatusErrorRuntimeFailure,
+                           "Fail to trying to Close Model");
+    }
+    return litert::Error(kLiteRtStatusErrorRuntimeFailure,
+                         "EnnAllocateAllBuffers Failed");
+  }
+
+  device_context->SetEnnCommittedBuffer(tmp_buf_ptr);
```
jl45G, line 122: shouldn't this be saved in invocation context instead of device context?

hoon98-choi: > shouldn't this be saved in invocation context instead of device context?

Thanks for the review! The commit buffer was created to be shared between models at the device level, but it feels more natural to have it in the invocation context. I’ll set that up in the second PR.

jl45G: Thanks! Resolve this comment

litert/tools/run_model.cc
Jiseong-oh: I just update CMake with samsung_flags.

jl45G (approved):

PR #6529, opened 2026-03-23
Qualcomm AI Engine Direct - Use median in run_model.

ChunHsuehLee:


litert/tools/run_model.cc
```diff
@@ -480,10 +457,14 @@ Expected<void> RunModel() {
...
-                 << std::accumulate(timers.begin(), timers.end(), uint64_t{0}) /
-                        timers.size()
-                 << " microseconds";
+  auto mid = timers.begin() + timers.size() / 2;
+  std::nth_element(timers.begin(), mid, timers.end());
+  const uint64_t median = timers.size() % 2 == 1
+                              ? timers[timers.size() / 2]
+                              : (*std::max_element(timers.begin(), mid) +
+                                 timers[timers.size() / 2]) /
+                                    2;
```
weilhuan-qti, line 466: I think we can sort `timers` to simplify implementation here.

BTW, do we want to merge this PR? or it is experimental.

litert/tools/run_model.cc
qti-hungjuiw, line 1: I think we can still keep the result of `average`.

PR #6532, opened 2026-03-23
Add to extra ops for Exynos Delegate

Jiseong-oh:
This is an addition to the ops supported by Exynos Exynos Delegate.
These ops have added basic NN operations, activation-related ops, and element-wise-related ops.


litert/vendors/samsung/compiler/builders/utils.cc
```diff
@@ -0,0 +1,94 @@
...
+    case ElementType::Bool:
+      return "BOOL";
+    case ElementType::Int4:
+      return "INT4";
+    case ElementType::Int8:
+      return "INT8";
+    case ElementType::UInt8:
+      return "UINT8";
+    case ElementType::Int16:
+      return "UINT16";
```
jl45G: is this a typo?

jl45G (approved):

litert/vendors/samsung/compiler/builders/utils.cc
Jiseong-oh: Yes, I fixed it.

PR #6633, opened 2026-03-28
[Intel OV] Add SHLOComposite op support to compiler plugin

akodanka:
This change adds support for kLiteRtOpCodeShloComposite, and currently supports the following attributes(for odml.rms_norm):
  - 'composite_name': via LiteRtGetSHLOCompositeOpName
  - 'epsilon': via GetOptionsAs<CompositeOptions>, with 1e-6 default

### DIFF
--- a/litert/vendors/intel_openvino/compiler/decoder.cc
+++ b/litert/vendors/intel_openvino/compiler/decoder.cc
@@ -32,6 +32,7 @@
 #include "litert/cc/litert_element_type.h"
 #include "litert/cc/litert_expected.h"
 #include "litert/cc/litert_macros.h"
+#include "litert/cc/internal/litert_op_options.h"
 #include "litert/vendors/intel_openvino/utils.h"
 #include "tflite/schema/schema_generated.h"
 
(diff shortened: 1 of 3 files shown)


litert/vendors/intel_openvino/compiler/decoder.cc
```diff
@@ -842,6 +843,30 @@ litert::Expected<ov::Any> DecoderOperation::fetch_attribute(
...
+            ERROR_LOG_STR("composite_op_name", op_name_.c_str()));
+        return ov::Any(std::string(composite_op_name));
+      } else if (name == "epsilon") {
+        auto info = GetOptionsAs<CompositeOptions>(litert_op_);
+        if (!info) {
+          return litert::Unexpected(kLiteRtStatusErrorRuntimeFailure,
+                                    "Failed to get composite options for " +
+                                        op_name_);
+        }
+        double epsilon = 1e-6; // Default value as per OV spec.
```
anikulk, line 860: Can these constants be defined with variable name like const kOVepsilon in decoder.h

anikulk (commented):
LGTM, just a minor nit

PR #6730, opened 2026-04-03
[Not for merge] Convert i2 weight to u2 for FC

taste1981:


anikulk:
Can you change this so prior to graph conversion tflite buffers are converted to uint2. So then we will map to ov_element_type u2

PR #6460, opened 2026-03-18
Refactor RemoteTensorBuffer to generic OpenVino tensor

zlma7001:
This change reduces direct dependency on Level Zero internal calls
and leverages OpenVINO's built-in memory management. By using the
recommended generic API, code readability and maintainability are
improved within the delegate while keeping the functional
behavior unchanged.

Test: .\run_model.exe --accelerator=npu --intel_openvino_device_type=npu
    --compiler_plugin_library_dir=... --dispatch_library_dir=... --graph=..

PR #6748, opened 2026-04-06
[Intel OV] Enable litert execution on linux

akodanka:
Enable litert execution across all OS using the Openvino buffer.

Verified the compiler plugin with the following steps :

Bazel build commands :
build litert/vendors/intel_openvino/compiler:libLiteRtCompilerPlugin_IntelOpenvino.so
build litert/vendors/intel_openvino/dispatch:libLiteRtDispatch_IntelOpenvino.so
build litert/tools:apply_plugin_main

AOT compilation step:
LD_LIBRARY_PATH=<PATH_TO_OPENVINO_LIBS> bazel-bin/litert/tools/apply_plugin_main --cmd apply --model=<PATH_TO_tflite_MODEL> -o <PATH_TO_AOT_MODEL> --libs bazel-bin/litert/vendors/intel_openvino/compiler --intel_openvino_device_type=cpu --soc_manufacturer "IntelOpenVINO" --soc_model=""

### DIFF
--- a/litert/vendors/intel_openvino/dispatch/BUILD
+++ b/litert/vendors/intel_openvino/dispatch/BUILD
@@ -27,10 +27,14 @@ litert_dynamic_lib(
         "device_context.cc",
         "dispatch_api.cc",
         "invocation_context.cc",
+        "openvino_shared_core.cc",
+        "openvino_tensor_buffer.cc",
     ],
     hdrs = [
         "device_context.h",
         "invocation_context.h",
+        "openvino_shared_core.h",
+        "openvino_tensor_buffer.h",
     ],
     copts = [
         "-Os",
--- a/litert/vendors/intel_openvino/dispatch/device_context.cc
+++ b/litert/vendors/intel_openvino/dispatch/device_context.cc
@@ -39,9 +39,7 @@
 #include "litert/c/litert_model.h"
 #include "litert/vendors/intel_openvino/utils.h"
 
-#if defined(LITERT_WINDOWS_OS)
 #include "litert/vendors/intel_openvino/dispatch/openvino_tensor_buffer.h"
-#endif  // LITERT_WINDOWS_OS
 
 #if LITERT_HAS_AHWB_SUPPORT || LITERT_HAS_DMABUF_SUPPORT
 #include "openvino/core/type/element_type.hpp"
@@ -143,7 +141,6 @@ LiteRtDispatchDeviceContextT::RegisterTensorBuffer(
 
   switch (tensor_buffer_type) {
     case kLiteRtTensorBufferTypeOpenVINOTensorBuffer: {
-#if defined(LITERT_WINDOWS_OS)
       HwMemoryHandle hw_memory_handle;
       LITERT_RETURN_IF_ERROR(
           LiteRtGetTensorBufferCustomTensorBufferHandle(tensor_buffer,
@@ -159,11 +156,6 @@ LiteRtDispatchDeviceContextT::RegisterTensorBuffer(
                                  RegisteredTensor{.tensor = openvino_tensor});
 
       return next_handle_++;
-#else
-      return litert::Unexpected(
-          kLiteRtStatusErrorRuntimeFailure,
-          "OpenVINO tensor support is missing on this platform.");
-#endif  // LITERT_WINDOWS_OS
     }
     case kLiteRtTensorBufferTypeDmaBuf: {
 #if LITERT_HAS_DMABUF_SUPPORT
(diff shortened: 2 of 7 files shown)


anikulk (approved):

litert/vendors/intel_openvino/dispatch/device_context.cc
```diff
@@ -52,63 +45,6 @@ LiteRtDispatchDeviceContextT::Create() {
   return Ptr(new LiteRtDispatchDeviceContextT());
 }
 
-#if LITERT_HAS_AHWB_SUPPORT
```
MatthiasKreileder, line 55: The goal is to be able to build LiteRT Intel compiler plugin / dispatch delegate on Linux, right?

Did you remove AHWB support because `LITERT_HAS_AHWB_SUPPORT` was enable to that build and gave errors?

MatthiasKreileder (changes requested):
Thanks for the work, I suppose now the Intel codepaths can be built for regular Linux, not just android_x86_64 and android_arm?

If so, would it be possible to add a test, given that we also have device type CPU?

litert/vendors/intel_openvino/dispatch/device_context.cc
akodanka: LITERT_HAS_AHWB_SUPPORT is enabled only for Android builds.
I removed it because it currently relies on creating ov::Tensor instead of RemoteTensor, thus RegisteredTensor will have to retain 2 different tensor types.
Additionally, it is unused currently and the implementation uses a workaround to fetch the buffer fd using socket APIs.

MatthiasKreileder: >  LITERT_HAS_AHWB_SUPPORT is enabled only for Android builds

Could we then keep this pull request only for changes to make it compile on Linux?

> I removed it because it currently relies on creating ov::Tensor instead of RemoteTensor, thus RegisteredTensor will have to retain 2 different tensor types.

And that would complicate to make it work on Linux?

In my simple mind only the change in litert/vendors/intel_openvino/dispatch/BUILD should be needed for Linux support.

litert/vendors/intel_openvino/dispatch/BUILD
```diff
@@ -54,8 +58,10 @@ litert_dynamic_lib(
     features = ["-use_header_modules"],  # Incompatible with -fexceptions.
     linkopts = litert_android_linkopts() + [
         "-Wl,-soname=libLiteRtDispatch_IntelOpenvino.so",
-        "-Wl,-lc++abi",
-    ],
+    ] + select({
```
MatthiasKreileder: Could you help me understand why another `select` is needed?

For other vendors we just need to append it:

https://github.com/google-ai-edge/LiteRT/blob/ceaf89c65502a698f2a368cfabc1cb2a8a65183d/litert/vendors/mediatek/dispatch/BUILD#L46
and https://github.com/google-ai-edge/LiteRT/blob/ceaf89c65502a698f2a368cfabc1cb2a8a65183d/litert/vendors/qualcomm/dispatch/BUILD#L50

litert/vendors/intel_openvino/dispatch/BUILD
MatthiasKreileder, line 25: With this change would it be possible to include building / running a test with Intel OpenVino in github CI https://github.com/google-ai-edge/LiteRT/blob/ceaf89c65502a698f2a368cfabc1cb2a8a65183d/.github/workflows/linux_x86_64.yml#L122

I guess we'd have to use a 'workaround' by passing in an option to the compiled model to use OpenVino plugin but with CPU, right? While not the production use-case it would be very helpful to automatically build and run the C++ code (for NPU execution we'll need device/lab based tests too, but those we don't have yet because of limited device availability).

MatthiasKreileder: If this isn't possible then please paste a short section in the pull request description of the test commands about you verified this change. Thanks for the work!

litert/vendors/intel_openvino/dispatch/device_context.cc
akodanka: I'll refactor this PR to make it specific to Linux.
So, for enabling on linux, we have to rely on the Openvino Tensor Buffer, which is already implemented for Windows. And, Android can also reuse that implementation, thus having a unified solution for all platforms.

MatthiasKreileder (commented):
Now with `PR #6460: Refactor RemoteTensorBuffer to generic OpenVino tensor` being merged https://github.com/google-ai-edge/LiteRT/pull/6855 please rebase to head.

Thanks!

MatthiasKreileder (approved):
Thanks!

MatthiasKreileder (approved):

MatthiasKreileder (approved):

PR #6824, opened 2026-04-10
Convert weight from i2 to u2 since OV does not support i2

taste1981:
OpenVINO does not support i2 data type. Thus before model compilation, we will need to convert the quantized i2 weights to use u2 data type, and shift zero-point accordingly, so they are numerically equivalent to the original i2 values.


MatthiasKreileder (approved):
Thanks!

Ideally we'd express model transformations via a LiteRT transformation https://github.com/google-ai-edge/LiteRT/blob/main/litert/vendors/examples/example_transformations_test.cc - but I don't want to hold up the merge of this change.

PR #6958, opened 2026-04-17
[intel_openvino] Add EliminateMatMulFakeQuantize pass to compiler

akodanka:
Implement a MatcherPass that removes FakeQuantize nodes following MatMul operations. This feature can be controlled during aot compilation using: --intel_openvino_configs_map="optimize_fq_after_matmul=true"

### DIFF
--- a/litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
+++ b/litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
@@ -24,9 +24,16 @@
 
 #include "openvino/core/any.hpp"
 #include "openvino/core/except.hpp"
+#include "openvino/core/graph_util.hpp"
+#include "openvino/core/rt_info.hpp"
 #include "openvino/frontend/tensorflow_lite/frontend.hpp"
 #include "openvino/frontend/tensorflow_lite/graph_iterator.hpp"
+#include "openvino/op/fake_quantize.hpp"
+#include "openvino/op/matmul.hpp"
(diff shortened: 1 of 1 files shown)


MatthiasKreileder (commented):
Thanks for the idea / suggestion.

1. Do you have a benchmark numbers for this optimization?

2. Could you outline briefly why we can throw away the FakeQuantize safely? Is it because the Intel NPU (OpenVino) compiler handles the quantization?

3. Minor: The team might ask if this can be done via a LiteRT transformation https://github.com/google-ai-edge/LiteRT/blob/f17ab20fb4dfa37e98a9e7ea8717aecb41854ee9/litert/compiler/plugin/compiler_plugin.h#L223 instead of OV transformation, to keep the code across vendors more similar.

akodanka:
> 1. Do you have a benchmark numbers for this optimization?
- We had observed 6-7% improvement in certain models.

> 2. Could you outline briefly why we can throw away the FakeQuantize safely? Is it because the Intel NPU (OpenVino) compiler handles the quantization?
- The output of the MatMul operation(NPU-optimized) in the OpenVINO IR is already FP32. An additional FQ layer will require further quantization(non-optimized) with the new quant parameters. 

> 3. Minor: The team might ask if this can be done via a LiteRT transformation
- This is primarily due to the quantization logic used in the OpenVINO+NPU, which introduces FakeQuantize layers for the Quantized input/output tensors, and I believe, it cannot be achieved by the LiteRT tranformations

PR #6975, opened 2026-04-19
Add device context and tensor identification params to custom tensor buffer handlers

devathul:
The current custom tensor buffer handler signatures (LiteRtTensorBufferCreateFunc) provide no device context, model information, or tensor index parameters, making it impossible to assign hardware-specific memory addresses during buffer creation. 

LiteRT's LiteRtRegisterTensorBufferHandlers API provides no way to determine:

1. Which device context does this buffer belong to 
2. Which model will this buffer be used with
3. Which tensor index (input[0], input[1], output[0], etc.) this buffer represents
4. What hardware memory address should be assigned

Proposing an enhanced LiteRtTensorBufferCreateFunc API signature that includes LiteRtDispatchDeviceContext, tensor_index, and is_input parameters, enabling vendors to achieve zero I/O tensor copy.
Since LiteRT invokes CreateBstmTensorBuffer sequentially for every input and output, there is no deterministic way to map a specific call to its corresponding pre-allocated hardware payload. Without identifying metadata, assigning the correct hardware memory addresses to the appropriate tensor buffers becomes an impossible mapping problem.

This change will avoid using global variables to pass LiteRtDispatchDeviceContext


devathul (commented):
This patch introduces three parameters to the custom buffer handler signatures to enable deterministic hardware memory mapping:

LiteRtDispatchDeviceContext: Provides the hardware-specific context directly to the handler
tensor_index: Identifies the specific position within the tensor array.
is_input: Disambiguates between input and output tensor arrays.

Changes:
API Update: Enhanced CreateCustomTensorBuffer and ImportCustomTensorBuffer in litert_custom_tensor_buffer.h with new context and identification parameters.

New Public API: Added LiteRtCompiledModelCreateBufferForIoTensor to the C API and runtime to allow explicit buffer creation for named I/O tensors.

Runtime Integration: Updated DispatchDelegateKernel to track tensor IDs and ExternalLiteRtBufferContext to propagate metadata during the allocation flow.

Added CreateManagedTensorBufferWithContext and updated CustomBuffer::Alloc to support context-aware allocation.

devathul:
@jl45G

PR #6976, opened 2026-04-20
Additional checks and Validation for Intel OV plugins

akodanka:
Introduced additional validation checks and fixed test executions.
This has changes on top of PR https://github.com/google-ai-edge/LiteRT/pull/6748 .


litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -198,13 +206,20 @@ litert_test(
...
+        "@intel_openvino//:openvino_libs",
     ],
+    env = select({
+        "//litert:linux": {
+            "LD_LIBRARY_PATH": "external/intel_openvino/openvino/runtime/lib/intel64",
+        },
+        "//conditions:default": {},
+    }),
     features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    linkstatic = 1,  # Avoid std::filesystem ABI conflicts with OpenVINO
```
MatthiasKreileder, line 218: For my education: Could you explain this error a bit more.

litert/vendors/intel_openvino/dispatch/dispatch_api_openvino_test.cc
```diff
@@ -83,219 +73,104 @@ TEST(OpenVino, DispatchApi) {
             kLiteRtStatusOk);
   EXPECT_NE(device_context, nullptr);
 
-  auto model_file_name =
```
MatthiasKreileder, line 86: What made you delete this? Is it because this relies running a device with actual Intel NPU?

MatthiasKreileder (commented):
Thanks!

litert/vendors/intel_openvino/compiler/BUILD
akodanka: Without linkstatic = 1, bazel builds separate shared libraries (.so files) for each dependency. 
The test binary ends up dynamically linking against libcommon.so from LiteRT and libopenvino.so 

When compiled as shared lib, the filesystem symbols in litert/test/common.cc seems to be exported and globally visible.
So, the std::filesystem usage from libopenvino.so ends up using this which causes a crash, as they were compiled with different C++ standard library versions.

With static linking, the std::filesystem symbols from LiteRT are hidden, and the corresponding calls from libopenvino.so gets correctly resolved.

litert/vendors/intel_openvino/dispatch/dispatch_api_openvino_test.cc
akodanka: Actually, the file never existed, it was added as a TODO.
If required, we need to plan introducing such a model for testing, but it might be a platform/device specific model.

I removed the model usage as this shouldn't block us from running the tests introduced.

MatthiasKreileder (approved):

PR #7121, opened 2026-04-28
Add support of a few TFL ops to OV compiler plugin.

taste1981:
Add support of ReduceAll and FloorMod, and Sign with limited float point support.


MatthiasKreileder (approved):
Thanks!

PR #7159, opened 2026-04-29
Add Bool support in npu_numerics_check output comparison

enlenama:
Previously, the get_val lambda had no handler for ElementType::Bool, this caused Bool-typed output tensors to remain as zero-initialized, producing misleading all-zero comparison results. Add a Bool handler that reads the data as uint8_t and converts to float for comparison.


enlenama:
Hi @MatthiasKreileder, could you please help review this change?

MatthiasKreileder (approved):

MatthiasKreileder (approved):

MatthiasKreileder (approved):

PR #7180, opened 2026-04-30
Restore SB function-name override [reversion] (b/397771624)

bpkeene:
The function-name override workaround in dispatch_api.cc was removed in e222c3c96 ("Update Google Tensor LiteRT for multi-signature support"), which exposed a backward-compatibility regression: SouthBound implementations shipped to date do not yet honor non-empty function names and reject any such call at thrInvocationContextGet time with

  E native: Failed to initialize the invocation context.
            INVALID_ARGUMENT: Only Darwinn Tflite model is supported for
            TpuTachyonLoadedLibrary.; node id: node_0,
            function_name: Partition_0

Reproducible on a stock Pixel 8 (Tensor G3) running an AOT-compiled  asset through LiteRtDispatchInvocationContextCreate.

### DIFF
--- a/litert/vendors/google_tensor/dispatch/dispatch_api.cc
+++ b/litert/vendors/google_tensor/dispatch/dispatch_api.cc
@@ -180,6 +180,14 @@ LiteRtStatus InvocationContextCreate(
   GT_LOG_RETURN_IF_NULL(exec_bytecode_buffer);
   GT_LOG_RETURN_IF_NULL(invocation_context);
 
+  // TODO(b/397771624): SouthBound implementations shipped to date do not
+  // support per-function dispatch and reject any non-empty function name
+  // ("Only Darwinn Tflite model is supported for TpuTachyonLoadedLibrary").
+  // Maintain the original empty-string override for older-SB compatibility.
+  // Once an SB version that supports per-function dispatch ships, gate this
+  // override behind a corresponding GoogleTensorSouthBoundFeature in
+  // sb_api_features.h.
+  function_name = "";
   return LiteRtDispatchInvocationContextT::CreateFromBytecode(
       device_context, exec_type, *exec_bytecode_buffer, function_name,
       num_inputs, num_outputs, *invocation_context);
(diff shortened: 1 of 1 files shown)


bpkeene:
cc @jl45G @tywerbicki-cmd verified on local devices that this restores eTPU inference on existing OS.  I suppose an alternate fix would be the AOT compiler generating bytecode that has a public function name?

bpkeene:
I'd very much appreciate guidance on proper version gating (SB API?) if possible.  Current PR simply reverts to maintain backwards compatibility.  Unfortunately this would break the multi-signature support from `e222c3c` - so I don't think the simple reversion should be merged as-is.

tywerbicki-cmd:
@bpkeene thank you for making me aware.

I've submitted a fix internally, and now waiting for copybara to make it publicly available. After which, this issue should be resolved.

As you're already familiar with, we're committed to maintaining backward compatibility with SouthBound shared objects via _sb_api_features.h_. In this case, we experienced a process failure, and have since taken actions to prevent such a failure from re-occurring.

Are you able to share the major and minor versions of the SouthBound shared object in your vendor image?

bpkeene:
Yes, we see-
G3: 0.9.0 
G4: 0.14.0
G5: 0.9.0

Curiously, G3 & G4 has sb_api available in 'libedgetpu_util.so' while G5 has the same in libedgetpu_litert.so

bpkeene:
sounds good! Can we keep this PR open until the official fix w/ a tagged release is shipped?

we'll use a fork with the patch until then.

tywerbicki-cmd:
@bpkeene thanks!

It looks like the fix landed as a part of https://github.com/google-ai-edge/LiteRT/commit/d6beffc83fcb18fee6304c9a59d5c4f58e869151. Can you please confirm that this mitigates the breakage you were observing?

```
Curiously, G3 & G4 has sb_api available in 'libedgetpu_util.so' while G5 has the same in libedgetpu_litert.so
```

Yes, this is a historical artifact. SouthBound symbols were originally exported by `libedgetpu_util.so`, and have since been moved to `libedgetpu_litert.so`.

PR #7253, opened 2026-05-06
[Intel OV] Add CPU backend support

daijh:
Extend Intel OpenVINO integration to support CPU execution.

- Implement seekoff and seekpos in CustomOStreamBuf, required by the CPU path.
- Use device-specific tensor allocation: 
  - NPU/GPU: create_host_tensor 
  - CPU: ov::Tensor
- Remove AUTO device and require explicit device targets (NPU, GPU, or CPU), because tensor allocation depends on the selected device type.
- Cache the remote context in OpenVINOSharedCore to prevent redundant creation.

Test:
bazel-bin\litert\tools\run_model.exe --print_tensors --compare_numerical --accelerator=npu --compiler_plugin_library_dir=%PLUGIN_DIR% --dispatch_library_dir=%PLUGIN_DIR% --graph=model.tflite --intel_openvino_device_type=cpu

### DIFF
--- a/litert/vendors/intel_openvino/compiler/openvino_compile_context.cc
+++ b/litert/vendors/intel_openvino/compiler/openvino_compile_context.cc
@@ -58,8 +58,8 @@ OpenVinoCompileContext::OpenVinoCompileContext() {
       context.device_ = "NPU";
       break;
     case kLiteRtIntelOpenVinoDeviceTypeAUTO:
-      context.device_ = "AUTO";
-      break;
+      return litert::Error(kLiteRtStatusErrorRuntimeFailure,
+                           "AUTO device type is not supported");
   }
 
   LITERT_LOG(LITERT_INFO, "Using Intel OpenVINO device: %s",
--- a/litert/vendors/intel_openvino/compiler/openvino_compile_context.cc
+++ b/litert/vendors/intel_openvino/compiler/openvino_compile_context.cc
@@ -57,9 +57,6 @@ OpenVinoCompileContext::OpenVinoCompileContext() {
     case kLiteRtIntelOpenVinoDeviceTypeNPU:
       context.device_ = "NPU";
       break;
-    case kLiteRtIntelOpenVinoDeviceTypeAUTO:
-      return litert::Error(kLiteRtStatusErrorRuntimeFailure,
-                           "AUTO device type is not supported");
   }
 
   LITERT_LOG(LITERT_INFO, "Using Intel OpenVINO device: %s",
--- a/litert/vendors/intel_openvino/dispatch/openvino_tensor_buffer.cc
+++ b/litert/vendors/intel_openvino/dispatch/openvino_tensor_buffer.cc
@@ -39,14 +39,23 @@ litert::Expected<void> OpenVinoTensorBuffer::Alloc(
   // TODO:: Release the shared OpenVINO Core.
   std::shared_ptr<ov::Core> core = OpenVINOSharedCore::GetInstance()->getCore();
   std::string device = OpenVINOSharedCore::GetInstance()->GetDevice();
-  auto context = core->get_default_context(device);
   ov::element::Type ov_element_type =
       litert::openvino::MapLiteTypeToOV(tensor_type.element_type);
+
   std::vector<int32_t> ov_shape_vec(tensor_type.layout.rank);
   for (size_t i = 0; i < ov_shape_vec.size(); i++)
     ov_shape_vec[i] = tensor_type.layout.dimensions[i];
-  host_tensor_ = context.create_host_tensor(
-      ov_element_type, ov::Shape{ov_shape_vec.begin(), ov_shape_vec.end()});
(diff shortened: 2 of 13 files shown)


daijh:
@MatthiasKreileder @anikulk PTAL

daijh:
@MatthiasKreileder @anikulk a friendly ping, PTAL.

MatthiasKreileder (changes requested):
Thanks for the work!

Could you please add an automated test that uses the cpu device type so it runs this as part of the github PR checks?

litert/vendors/intel_openvino/dispatch/openvino_tensor_buffer.cc
```diff
@@ -39,14 +39,23 @@ litert::Expected<void> OpenVinoTensorBuffer::Alloc(
...
+
   std::vector<int32_t> ov_shape_vec(tensor_type.layout.rank);
   for (size_t i = 0; i < ov_shape_vec.size(); i++)
     ov_shape_vec[i] = tensor_type.layout.dimensions[i];
-  host_tensor_ = context.create_host_tensor(
-      ov_element_type, ov::Shape{ov_shape_vec.begin(), ov_shape_vec.end()});
+  ov::Shape ov_shape{ov_shape_vec.begin(), ov_shape_vec.end()};
+
+  if (device == "NPU" || device == "GPU") {
+    auto context = core->get_default_context(device);
```
akodanka: Can we reuse the context instead of creating a new context for each buffer allocation ?

litert/vendors/intel_openvino/compiler/openvino_compile_context.cc
```diff
@@ -58,8 +58,8 @@ ::litert::Expected<OpenVinoCompileContext> OpenVinoCompileContext::Create(
       context.device_ = "NPU";
       break;
     case kLiteRtIntelOpenVinoDeviceTypeAUTO:
-      context.device_ = "AUTO";
```
akodanka, line 61: In that case, Would it be better to remove the support for "AUTO" itself, instead of having a conditional failure here ?

litert/vendors/intel_openvino/dispatch/openvino_tensor_buffer.cc
daijh: Thanks @akodanka
My understanding is that, internally, `get_default_context()` caches the handle and returns it directly if it already exists, meaning the caller does not need to maintain a separate cache.

daijh: > My understanding is that, internally, get_default_context() caches the handle and returns it directly if it already exists, meaning the caller does not need to maintain a separate cache.

Ah, got it! So `get_default_context()` does not actually manage the lifetime or caching for you under the hood.
Let me fix it.

daijh: Fixed

litert/vendors/intel_openvino/compiler/openvino_compile_context.cc
daijh: Great suggestion, fixed.

daijh:
> Could you please add an automated test that uses the cpu device type so it runs this as part of the github PR checks?

Sure.

It appears that the C++ tests are currently failing on the main branch(8d715c2c0), independent of this PR.
May I defer the CPU device testing updates to a subsequent PR?

```shell
bazelisk build --config=windows //litert/vendors/intel_openvino/compiler:openvino_compiler_plugin_test
```

MatthiasKreileder:
> > Could you please add an automated test that uses the cpu device type so it runs this as part of the github PR checks?
> 
> Sure.
> 
> It appears that the C++ tests are currently failing on the main branch([8d715c2](https://github.com/google-ai-edge/LiteRT/commit/8d715c2c0309a65c2e69e9c15b4828756164fe2e)), independent of this PR. May I defer the CPU device testing updates to a subsequent PR?
> 
> ```shell
> bazelisk build --config=windows //litert/vendors/intel_openvino/compiler:openvino_compiler_plugin_test
> ```

It's ok to ignore the existing failures.

But please add a new test driver as a separate target then, make that uses OpenVino CPU and and ensure that the github check runs this test successfully on windows and linux. This would really help with coverage.

daijh:
The `openvino_compiler_plugin_test` and `dispatch_api_openvino_test` targets currently fail to build:

```
bazelisk build --config=windows //litert/vendors/intel_openvino/compiler:openvino_compiler_plugin_test
bazelisk build --config=windows //litert/vendors/intel_openvino/dispatch:dispatch_api_openvino_test
```

These are tracked in issue #7878. Note that neither target is run in GitHub checks — both are explicitly excluded from CI and tagged manual. Once the build issues are resolved, we will add OpenVINO CPU test cases as a separate target.

All other OpenVINO-related targets (plugin builds, options tests, flag tests) pass on both Windows and Linux CI.

MatthiasKreileder (approved):

PR #7257, opened 2026-05-06
[Intel OV] Python npu compiler fix

rairatne:
*Bundle Intel NPU compiler in SDK sdist + auto‑discover dispatch path**

Fixes Intel NPU AOT compilation failures caused by the OpenVINO pip wheel missing
the NPU compiler library. The Intel SDK now downloads
`libopenvino_intel_npu_compiler.{so,dll}` at pip install time from the official
OpenVINO toolkit, keeping the sdist small and matching the Qualcomm SDK flow.

**Highlights**
- OS‑aware archive selection (ubuntu22 vs ubuntu24; Windows zip)
- Non‑fatal, skippable downloads for hermetic builds
- Automatic DLL resolution on Windows and correct library propagation to AOT
- Auto‑discovery of `dispatch_library_path` (explicit paths still win)
- CI guard enforcing OpenVINO version pin sync to prevent ABI drift

**Result**
Intel NPU AOT compile and `litert-benchmark --use_npu` work out of the box.

### DIFF
--- a/ci/check_openvino_version_sync.py
+++ b/ci/check_openvino_version_sync.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+"""Fail CI when OpenVINO versions drift between setup.py and openvino.bzl.
+
+The Intel OV compiler plugin and dispatch library are built against the
+OpenVINO SDK pinned in third_party/intel_openvino/openvino.bzl. The NPU
+compiler shared library is fetched by ci/tools/python/vendor_sdk/intel/setup.py
+at pip install time. Both must pin the same OpenVINO build.
+"""
+
+import pathlib
+import re
+import sys
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parent.parent
+BZL = REPO_ROOT / "third_party/intel_openvino/openvino.bzl"
+SETUP = REPO_ROOT / "ci/tools/python/vendor_sdk/intel/setup.py"
+
+# OpenVINO build identifier: "<year>.<minor>.<patch>.<build>.<hash>", e.g.
+# "2026.1.0.21367.63e31528c62". The trailing commit hash is 7-40 hex chars.
+# The surrounding context in openvino.bzl URLs ("..._x86_64.tgz") is not a
+# word boundary for Python's \b, so we match a 5-component dotted identifier
+# ending with the hash and rely on the preceding "/" or "_" being non-alnum.
+_BUILD_RE = re.compile(r"\d{4}\.\d+\.\d+\.\d+\.[0-9a-f]{7,40}")
+_SETUP_BUILD_RE = re.compile(r"_OV_BUILD\s*=\s*'([^']+)'")
+
+
+def main() -> int:
+  bzl_builds = set(_BUILD_RE.findall(BZL.read_text()))
+  m = _SETUP_BUILD_RE.search(SETUP.read_text())
+  if not m:
+    print(f"ERROR: could not find _OV_BUILD in {SETUP}", file=sys.stderr)
+    return 1
+  setup_build = m.group(1)
+
+  if len(bzl_builds) != 1:
+    print(
+        f"ERROR: {BZL} references multiple OpenVINO builds: {sorted(bzl_builds)}",
+        file=sys.stderr,
+    )
+    return 1
+  bzl_build = next(iter(bzl_builds))
+
+  if setup_build != bzl_build:
+    print(
+        "ERROR: OpenVINO build pin drift:\n"
+        f"  {BZL}: {bzl_build}\n"
+        f"  {SETUP}: {setup_build}\n"
+        "Both must pin the same OpenVINO release.",
+        file=sys.stderr,
+    )
+    return 1
+
+  print(f"OK: OpenVINO build {bzl_build} matches in both files.")
+  return 0
+
+
+if __name__ == "__main__":
+  sys.exit(main())
--- a/litert/vendors/intel_openvino/README.md
+++ b/litert/vendors/intel_openvino/README.md
@@ -17,22 +17,8 @@ limitations under the License. -->
 This directory contains the Intel OpenVINO integration for LiteRT, providing
 compiler plugin and dispatch API support for Intel NPU hardware.
 
-## Table of Contents
-
--   [Supported Platforms](#supported-platforms)
--   [Prerequisites](#prerequisites)
--   [Quick Start](#quick-start)
--   [Verifying NPU Execution](#verifying-npu-execution)
--   [AOT Compilation](#aot-compilation)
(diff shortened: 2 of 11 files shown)


litert/vendors/intel_openvino/README.md
```diff
@@ -43,70 +29,56 @@ Intel Core Ultra Series 3 | NPU5010 | Panther Lake (PTL) | Linux, Windows
...
------------ | ----------------------------------- | -------
-OS          | Ubuntu 24.04 LTS (x86_64)           | Windows 10/11 (x86_64)
-Python      | 3.10, 3.11, 3.12, or 3.13           | 3.11 only
-OpenVINO    | `pip install openvino==2026.1.0`    | `pip install openvino==2026.1.0`
-NPU driver  | v1.32.0 (system package via `dpkg`) | 32.0.100.4724+ ([download](https://www.intel.com/content/www/us/en/download/794734/intel-npu-driver-windows.html))
+Requirement | Linux                              | Windows
+----------- | ---------------------------------- | -------
+OS          | Ubuntu 22.04 or 24.04 LTS (x86_64) | Windows 10/11 (x86_64)
+Python      | 3.10 – 3.13                        | 3.11 only
+NPU driver  | v1.32.1 (system `.deb`)            | 32.0.100.4724+ ([download](https://www.intel.com/content/www/us/en/download/794734/intel-npu-driver-windows.html))
```
MatthiasKreileder: Thanks for the quick turnaround on this.

Could we please make it clear already here in which scenario the NPU driver installation is a requirement? You are mentioning this further down in the doc to differentiate between AOT and JIT, but would be good to say it here to avoid confusion.

IIUC we should have the following:

## AOT compiling a tflite model for Intel NPU

Regardless of the host OS (Linux / Windows) when performing AOT compilation, the NPU driver *does not need to be installed*. The user can `pip install ai-edge-litert-nightly ai-edge-litert-sdk-intel-nightly` and can then compile a model.

The resulting model can then be used for inference (on Linux or Windows). We should also call that "cross-compilation" is working, i.e. the compilation step can happen on Linux and then the pre-compiled model can be run on Windows. Which is cool :)

## tflite model inference Intel NPU

This seems to depend on the API being used for inference.

For users of the C++ API, NPU driver installation is *not* required, because LiteRT fetches the OpenVino prebuilt C++ SDK, which contains the required shared libs, and inference (either AOT or JIT) is working.

For users of the Python API, if I understand correctly, NPU driver installation is required, because the python package won't have all the required files?

ci/check_openvino_version_sync.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+"""Fail CI when OpenVINO versions drift between setup.py and openvino.bzl.
+
+The Intel OV compiler plugin and dispatch library are built against the
```
MatthiasKreileder, line 6: Thanks for adding a check.

Do I understand correctly that once the python package of openvino on Linux also contains the openvino compiler plugin we don't need to keep setup.py and openvino.bzl in sync?

I guess using an updated openvino version in openvino.bzl should be ok, say to run a model that has been compiled via the python package with an older openvino version, right?

No need to block on this now, just want to clarify.

MatthiasKreileder (commented):
Thanks, this looks good. Some minor comments on clarification.

ci/check_openvino_version_sync.py
rairatne: > setup.py will still be needed for ai-edge-litert-sdk-intel but this customization to download the openvino archive and pull the npu compiler will not be deprecated, it will go back to thin wrapper on top of openvino package like earlier.


> using an updated openvino version in openvino.bzl should be ok, say to run a model that has been compiled via the python package with an older openvino version
             Yes most of cases newer openvino version do supports AOT models compiled with older version (backward compatibility is planned to be supported from OpenVINO 2026.2 onwards)

litert/vendors/intel_openvino/README.md
rairatne: > For users of the Python API, if I understand correctly, NPU driver installation is required, because the python package won't have all the required files? 
     -> Yes for inference on NPU,  driver installation is needed
     

- I will update the README.md highlighting it properly

rairatne: Updated Readme

MatthiasKreileder (approved):

MatthiasKreileder (approved):

litert/vendors/intel_openvino/README.md
```diff
@@ -43,72 +29,93 @@ Intel Core Ultra Series 3 | NPU5010 | Panther Lake (PTL) | Linux, Windows
...
-[Bazelisk](https://github.com/bazelbuild/bazelisk)) - Docker (for hermetic
-builds)
+### When is the NPU driver required?
+
+| Task                                                | Needs NPU driver installed? |
+| --------------------------------------------------- | --------------------------- |
+| AOT compile a `.tflite` for Intel NPU               | **No.** Cross-compilation works — e.g. compile on Linux, run the resulting `.tflite` on Windows. |
+| Run inference on an Intel NPU (Python or C++ API, JIT or AOT-compiled model) | **Yes.** The dispatch library talks to the hardware via the Level Zero loader shipped in the NPU driver package. |
+
+> **In short:** the NPU driver is needed only on machines that **execute** the
```
MatthiasKreileder, line 45: What I still find confusing is that when I use the C++ API of LiteRT all I need is the openvino SDK, I don't need to separately install the NPU driver. So why is this needed for Python?

It would be great if we can simplify and avoid the separate installation of the NPU driver, as it's not needed at least for C++, unless I am missing something?

MatthiasKreileder (commented):
Please rebase, sadly github was out of sync with the latest changes.

litert/vendors/intel_openvino/README.md
MatthiasKreileder: Sorry also but the google -> github sync was temporarily not working so the environment.py was slightly out of date (fixed now). Please rebase and I'll LGTM again.

MatthiasKreileder: If you want to (to speed-up the review + import cycle) you could address the OpenVino library discovery issue for JIT-inference (longer write-up in https://docs.google.com/document/d/1tjI08D9ocw6_bs2cWEkyHDJreca4XSPL13uBronPDWQ/edit?resourcekey=0-3rzlIfCxNlGI3PCUaTQrAQ&tab=t.krg33h51yefq)

rairatne: Hi @MatthiasKreileder NPU Driver is needed, without it how is NPU getting listed? 

Are you talking about only AOT or inference too?
For inference how are checking if NPU is getting used at all and not falling back to  xnnpack

rairatne: Rebased

MatthiasKreileder (approved):

PR #7341, opened 2026-05-11
fix(apple): load GPU accelerators from framework bundles

mmaloney007:
## Summary
- add an Apple framework-binary fallback in GPU accelerator registration
- map basename loads such as `libLiteRtMetalAccelerator.dylib` to `<runtime_lib_path>/LiteRtMetalAccelerator.framework/LiteRtMetalAccelerator`
- add diagnostic logging around accelerator load/register failures

## Why
iOS/TestFlight apps cannot ship raw nested dylibs; accelerator sidecars need to be packaged as signed frameworks. The existing basename load path cannot find those framework binaries from the app sandbox, so GPU registration fails before compiled-model creation.

## Validation
- Used with LiteRT-LM runtime-library-dir forwarding in b4 iPhone 16 Pro validation.
- Device stderr showed `GPU Metal` registered and Gemma 4 E2B decode completed via Metal.
- The same branch produced clearer diagnostics when registration failed during earlier bundle-layout tests.

### DIFF
--- a/litert/runtime/accelerators/gpu_registry.cc
+++ b/litert/runtime/accelerators/gpu_registry.cc
@@ -42,6 +42,27 @@ namespace litert::internal {
 #define SO_EXT ".so"
 #endif
 
+#if defined(__APPLE__)
+std::filesystem::path AppleFrameworkBinaryPath(
+    const std::filesystem::path& runtime_lib_path,
+    absl::string_view plugin_path) {
+  if (runtime_lib_path.empty()) {
+    return {};
+  }
+  std::string framework_name(plugin_path);
+  if (framework_name.rfind("lib", 0) == 0) {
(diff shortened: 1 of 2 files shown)


mmaloney007:
I have now signed the Google CLA for this GitHub account. Could you please re-run/confirm the CLA check and take a look when convenient? This PR is a small Apple packaging/runtime-path fix for loading LiteRT GPU accelerator sidecars from framework bundles.

mmaloney007:
CLA is green now. Could someone take a look when convenient?

This is a small Apple packaging/runtime-path fix that unblocks LiteRT GPU accelerator sidecar resolution from framework bundles on iOS/macOS. Happy to adjust the shape if there is a preferred upstream direction.

PR #7298, opened 2026-05-08
Fix readme

rairatne:
Reformat and Cleanup Intel OV Readme

PR #7364, opened 2026-05-12
[Intel OV] Drop direct-URL openvino pin (PyPI rejects)

rairatne:
PyPI policy blocks Requires-Dist entries of the form `<name> @ <url>`, so the nightly upload job was 400ing on the Intel sdist:

  400 Can't have direct dependency: openvino @
  https://storage.openvinotoolkit.org/wheels/nightly/openvino/...

Switch install_requires to a plain PEP 440 pin
(`openvino==2026.2.0.dev20260506`) and put the `--extra-index-url` step back in the README install command. Delete the now-unused _OV_WHEEL_BASE_URL, _OV_WHEEL_TARGETS, and _openvino_install_requires() from setup.py.

The trade-off is small: pip picks whichever wheel with that date stamp has the newest build tag at install time. Same-date respins are rare, and check_openvino_version_sync.py still catches archive vs setup.py date-stamp drift.

Long-term, OpenVINO will bundle the NPU compiler library in the Linux PyPI wheel (matching the Windows wheel); at that point the toolkit archive fetch and the compiler-lib copy logic can be dropped entirely.

### DIFF
--- a/ci/tools/python/vendor_sdk/intel/setup.py
+++ b/ci/tools/python/vendor_sdk/intel/setup.py
@@ -49,16 +49,19 @@ IS_X86_ARCHITECTURE = platform.machine() in ('x86_64', 'AMD64', 'i386', 'i686')
 # --- Configuration for OpenVINO NPU Compiler Download ---
 #
 # Must match the OpenVINO build pinned in third_party/intel_openvino/
-# openvino.bzl. That workspace file pins the OpenVINO SDK used to compile the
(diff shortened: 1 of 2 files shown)


MatthiasKreileder (approved):
Thanks for the quick fix.

rairatne:
Added rebased Readme.md cherrypicked from https://github.com/google-ai-edge/LiteRT/pull/7298

PR #7484, opened 2026-05-19
[Github copilot]add clang-tidy GitHub Action

jl45G:
## Summary
- Adds a clang-tidy GitHub Actions workflow for pull requests to main.
- Configures a CMake compile database and runs clang-tidy on changed C/C++ source files covered by that database.
- Always includes `litert/c/empty.cc` as a smoke translation unit so workflow-only changes still execute clang-tidy.

## Validation
- Local: CMake configure generated `litert/cmake_build_clang_tidy/compile_commands.json`.
- Local: file selection chose `litert/c/empty.cc` for this workflow-only diff.
- Local: clang-tidy smoke passed on `litert/c/empty.cc` using the generated compile database.
- GitHub Actions: `clang-tidy` check passed on this PR: https://github.com/google-ai-edge/LiteRT/actions/runs/26069691174

After this workflow lands on `main`, the `clang-tidy` check name can be added to branch protection or the `LiteRT_main` ruleset as a required status check.

### DIFF
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -1,0 +1,148 @@
+name: "clang-tidy"
+
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  clang-tidy:
+    name: "clang-tidy"
+    runs-on: linux-x86-n2-16
+    container:
+      image: us-docker.pkg.dev/ml-oss-artifacts-published/ml-public-container/ml-build:latest
+    env:
+      CLANG_TIDY_CHECKS: "clang-diagnostic-*,clang-analyzer-*"
+      CMAKE_BUILD_DIR: "litert/cmake_build_clang_tidy"
+      GITHUB_TOKEN: ${{ github.token }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Install clang-tidy dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            clang-18 \
+            clang-tidy-18 \
+            libc++-18-dev \
+            libc++abi-18-dev \
+            python3-pip
+          pip3 install cmake==3.28.3
+          cmake --version
+          clang-tidy-18 --version
+
+      - name: Configure CMake compile database
+        run: |
+          cmake -S litert -B "${CMAKE_BUILD_DIR}" \
+            -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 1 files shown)


.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,148 @@
...
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Install clang-tidy dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            clang-18 \
+            clang-tidy-18 \
```
snnn, line 39: Suggest using the latest clang version to get better diagnosis

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,148 @@
...
+          cmake -S litert -B "${CMAKE_BUILD_DIR}" \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+            -DCMAKE_C_COMPILER=clang-18 \
+            -DCMAKE_CXX_COMPILER=clang++-18 \
+            -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
+            -DLITERT_BUILD_TESTS=OFF \
+            -DLITERT_ENABLE_GPU=OFF \
+            -DLITERT_ENABLE_NPU=OFF
+          test -f "${CMAKE_BUILD_DIR}/compile_commands.json"
```
snnn, line 58: There are two ways of doing this:
1. Generate a compile database then pass it to clang-tidy
2. Directly run clang-tidy in cmake by setting https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html

You may choose which is easier to setup

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,148 @@
...
+            else:
+              source = source.resolve()
+            try:
+              compilable.add(source.relative_to(repo).as_posix())
+            except ValueError:
+              pass
+
+          source_suffixes = {".c", ".cc", ".cpp", ".cxx"}
+          changed = []
+          event_name = os.environ["GITHUB_EVENT_NAME"]
```
snnn, line 87: Generate a *.sarif file instead. Then Github will handle this. 

See: https://docs.github.com/en/code-security/reference/code-scanning/codeql/codeql-cli/sarif-output

The file format is public. 

We shall avoid directly manipulate Github REST APIs with ACCESS_TOKEN, because it's easy to cause security leakages.

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,148 @@
+name: "clang-tidy"
+
```
snnn, line 2: You will also need to add a .clang-tidy file in the repo's root dir.

.github/workflows/clang_tidy.yml
snnn: Here is the task: https://github.com/github/codeql-action/blob/main/upload-sarif/action.yml (The sarif file does not need to be generated from CodeQL)

PR #7605, opened 2026-05-25
[Win]Fix FileCopyAllocation to use 64-bit fstat for large model query

enlenama:
On Windows `struct stat::st_size` is a 32-bit which overflows for files larger than 2GB. Use `_fstat64` with `struct __stat64` to correctly retrieve 64-bit file sizes.

### DIFF
--- a/tflite/converter/allocation.cc
+++ b/tflite/converter/allocation.cc
@@ -39,19 +39,17 @@ FileCopyAllocation::FileCopyAllocation(const char* filename,
     error_reporter_->Report("Could not open '%s'.", filename);
     return;
   }
-  struct stat sb;
-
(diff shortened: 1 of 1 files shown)


enlenama:
Hi @MatthiasKreileder, could you please help review this change?

MatthiasKreileder:
Sorry for being late here. We have this change now imported as part of https://github.com/google-ai-edge/LiteRT/pull/7925 I hope that's ok.

enlenama:
> Sorry for being late here. We have this change now imported as part of #7925 I hope that's ok.

Glad to see the changes made it in via #7925. As long as the fix has landed, it's all good. I will close this PR to keep things tidy.

PR #7698, opened 2026-05-31
Support specifying the OV backend for subgraph compilation/dispatch.

taste1981:
You can specify the backend used per-partition via its 0-based partition index for apply_plugin_main:
`apply_plugin_main.exe ... --intel_openvino_graph_backends="0:npu;1:gpu" ...`, or you can apply one backend  to
all the partitions by `--intel_openvino_graph_backends="gpu" ...`

Also remove the intel_openvino_device_type option when compiling the models, to converge to intel_openvino_graph_backends option instead. Subgraph without explicit backend defaults to NPU.


litert/vendors/intel_openvino/bytecode_header.h
```diff
@@ -0,0 +1,109 @@
...
+  OpenVinoBytecodeHeader hdr;
+  std::memcpy(&hdr, bytecode, sizeof(hdr));
+  if (std::memcmp(hdr.magic, OpenVinoBytecodeHeader::kMagic,
+                  sizeof(hdr.magic)) != 0) {
+    return false;
+  }
+  if (hdr.version != OpenVinoBytecodeHeader::kCurrentVersion) {
+    return false;
+  }
+  if (graph_backend != nullptr) {
```
anikulk, line 103: Can we add few gaurdrails around the hdr.graph_backend valid range. We can define a sentinal value : 
  kLiteRtIntelOpenVinoGraphBackendMax = 4 to LiteRtIntelOpenVinoGraphBackend and check if graph_backend is less than kLiteRtIntelOpenVinoGraphBackendMax .

litert/c/options/litert_intel_openvino_options.cc
```diff
@@ -68,15 +78,22 @@ LiteRtStatus LrtGetOpaqueIntelOpenVinoOptionsData(
...
   if (options->performance_mode.has_value()) {
     ss << "performance_mode = " << options->performance_mode.value() << "\n";
   }
   for (const auto& pair : options->configs_map_options) {
     ss << "configs_map." << pair.first << " = \"" << pair.second << "\"\n";
   }
+  for (const auto& [graph_index, override] : options->graph_overrides) {
+    if (override.graph_backend.has_value()) {
+      ss << "graph." << graph_index
+         << ".graph_backend = " << override.graph_backend.value() << "\n";
```
anikulk, line 90: Should we add a device availability check before using the device from the header?

anikulk (commented):
Thanks for the change, few comments.

litert/vendors/intel_openvino/bytecode_header.h
taste1981: done.

litert/c/options/litert_intel_openvino_options.cc
taste1981: done. Add device check in the dispatch.

MatthiasKreileder (approved):

MatthiasKreileder (approved):

litert/tools/flags/vendors/README_intel_openvino_flags.md
```diff
@@ -38,28 +31,58 @@ Flag                           | Type   | Default | Description
...
+# AOT-compile a 2-signature model: partition 0 on NPU, partition 1 on CPU
+# with a per-partition precision hint
+apply_plugin_main \
+  --cmd=apply \
+  --model=/path/to/model.tflite \
+  --soc_manufacturer=IntelOpenVINO \
+  --soc_model=PTL \
+  --libs=/path/to/plugin/dir \
+  --o=/path/to/output.tflite \
+  --intel_openvino_graph_backends="0:npu;1:cpu" \
```
MatthiasKreileder, line 84: Thanks for the PR, overall looks great and multi-accelerator support is good to have.

I've consulted with the team and one concern came up: Could we please use signature names instead of indices? At least in the "user-facing" APIs? One of the things that LiteRT wants to get right over tflite is that signature names are used instead of indices for improved readability and easier debugging.

MatthiasKreileder (changes requested):

litert/tools/flags/vendors/README_intel_openvino_flags.md
taste1981: One of the purpose of this to allow offloading some of the transformer layers to GPU while other layers on NPU, to maximize usage of the compute power for the platform.

With that you'll not be able to get a signature - the original sub-graph with signature (for example, decode or prefill-128) has already been either partitioned by LiteRT, or by other high-level components that enables pipelining.

It is different from the capability of `apply_plugin_main` which handles subgraph. This instead, handles partitions in a specific subgraph.

MatthiasKreileder: Thanks for clarifying, resolving this.

Issue #4079, opened 2025-10-17
ReduceSum reference kernel overflows when network type not uint8


idalevison:
**System information**
- Linux Ubuntu 22.04
- LiteRT installed from pip source
- LiteRT version 2.0.2 

**Standalone code to reproduce the issue** 

```
import numpy as np
from ai_edge_litert.interpreter import Interpreter
from ai_edge_litert.interpreter import OpResolverType

interpreter = Interpreter(model_path="minimal.tflite", experimental_op_resolver_type=OpResolverType.BUILTIN_REF)
interpreter.allocate_tensors()

inp = interpreter.get_input_details()[0]
out = interpreter.get_output_details()[0]

x = np.full(inp["shape"], 2, dtype=inp["dtype"])
interpreter.set_tensor(inp["index"], x)
interpreter.invoke()
y = interpreter.get_tensor(out["index"])
print(y)
```

The example network can be created from the attached minimal.json. The network contains a single ReduceSum Op, the input tensor has shape and type int8[1,128].

Output is [[-1]] when it should be [[127]]

[minimal.json](https://github.com/user-attachments/files/22972447/minimal.json)

<img width="244" height="423" alt="Image" src="https://github.com/user-attachments/assets/a25528b0-df9f-4d1f-80d6-0116abfc6576" />

**Any other info / logs**

I have located the bug in the tflite reference kernel, in file tflite/kernels/reduce.cc.
In the function QuantizedMeanOrSum, when kernel_type is kReference, the call to reference_ops::QuantizedMeanOrSum uses type uint8 instead of template type T.
This results in overflow when a network has a different type. I have included the bug fix as a git patch below.

```
diff --git a/tflite/kernels/reduce.cc b/tflite/kernels/reduce.cc
index 8ead4274826..0a7afb5a2a5 100644
--- a/tflite/kernels/reduce.cc
+++ b/tflite/kernels/reduce.cc
@@ -477,10 +477,10 @@ TfLiteStatus QuantizedMeanOrSum(TfLiteContext* context,
     TF_LITE_ENSURE(
         context,
         reference_ops::QuantizedMeanOrSum(
-            GetTensorData<uint8_t>(op_context.input),
+            GetTensorData<T>(op_context.input),
             op_context.input->params.zero_point, op_context.input->dims->data,
             op_context.input->dims->size,
-            GetTensorData<uint8_t>(op_context.output), op_data->multiplier,
[...]

PR #7855, opened 2026-06-04
TFLite: Fix QuantizedMeanOrSum template type in refkernel

idalevison:
Fixes ReduceSum overflow for non-uint8 tensors by passing the template type to GetTensorData in the QuantizedMeanOrSum.


idalevison:
Fixes issue #4079

PR #7931, opened 2026-06-08
runtime: allow auto-resize of input tensors without dims_signature

zlma7001:
InputTensorNeedsResize previously refused to auto-resize any tensor that lacked a dims_signature, which blocked LLM/KV-cache inputs exported without explicit dynamic-shape annotations.

Treat a missing or empty dims_signature as fully dynamic: accept the new shape when its rank matches the current shape and all dimensions are positive. Update the case-list comment accordingly and log an INFO message when this fallback path is taken.

### DIFF
--- a/litert/runtime/compiled_model.cc
+++ b/litert/runtime/compiled_model.cc
@@ -2064,8 +2064,11 @@ Expected<LiteRtMetricsT> LiteRtCompiledModelT::StopMetricsCollection() const {
 // needed.
 // 2. The input tensor has dynamic dimensions and the new shape is compatible,
 // then resize is needed.
-// 3. The input tensor has static dimensions or the new shape is not compatible,
-// then return error.
+// 3. The input tensor has no dims_signature but shapes differ: allow resize
+// (treat all dims as dynamic). This supports KV cache and other tensors in
+// LLMs that may lack explicit dynamic shape annotations.
+// 4. The input tensor has static dimensions and the new shape is not
+// compatible, then return error.
 Expected<bool> LiteRtCompiledModelT::InputTensorNeedsResize(
     const TfLiteTensor* tensor, absl::Span<const int> new_shape) {
   const TfLiteIntArray* shape_array = tensor->dims;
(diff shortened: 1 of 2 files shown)


zlma7001:
@MatthiasKreileder , could you please review this change? thanks!

MatthiasKreileder (changes requested):
Looks good! Please add a unit test.

zlma7001:
Thanks Matt, unit test added, PTAL

MatthiasKreileder (approved):

litert/runtime/compiled_model.cc
```diff
@@ -2084,10 +2086,32 @@ Expected<bool> LiteRtCompiledModelT::InputTensorNeedsResize(
...
 
   if (!tensor->dims_signature || tensor->dims_signature->size == 0) {
-    return Unexpected(kLiteRtStatusErrorInvalidArgument,
-                      absl::StrCat("Cannot auto-resize tensor ",
-                                   tensor->name ? tensor->name : "<unnamed>",
-                                   ": no dims_signature exists"));
+    // No dims_signature: allow resize if the rank matches and all new
+    // dimensions are positive.  This is the common case for KV-cache tensors
+    // in LLM models that were exported without explicit dynamic-shape
+    // annotations.
```
jl45G, line 2092: This logic is too broad, this broaddens auto-resize for any static input tensor without a shape signature, not just kv-cache tensors 

see the definition of dims_signature: https://developers.google.com/edge/api/tflite/c/struct/tf-lite-tensor#dims_signature

jl45G (changes requested):

jl45G:
This doesn't seem to be the right change, the definition of dims_signature is

"""
Encodes shapes with unknown dimensions with -1. This field is only populated when unknown dimensions exist in a read-write tensor (i.e. an input or output tensor). (e.g. dims contains [1, 1, 1, 3] and dims_signature contains [1, -1, -1, 3]). If no unknown dimensions exist then dims_signature is either null, or set to an empty array. Use TfLiteTensorGetDimsSignature to get dims_signature if non-empty or otherwise fallback to dims. Note that this field only exists when TF_LITE_STATIC_MEMORY is not defined.
"""

zlma7001:
Many thanks to @jl45G  and @MatthiasKreileder  for providing this information. I will take a deeper look at this issue `ERROR: [litert/c/litert_compiled_model.cc:156] Failed to register input tensor buffer: ERROR: [unknown:0]
Γöö Cannot auto-resize tensor kv_cache_k_5: no dims_signature exists`. Thanks.

zlma7001:
Thanks all again, this model run fails because the model declares inconsistent KV-cache shapes across signatures: prefill_128 and decode treat the global-attention layers (5, 11, 17, …) as [1, 4, 4095, 512], but prefill_cache_update_128 and decode_cache_update declares those same shared cache tensors as [1, 16, 4095, 256], not related to LiteRT, so closing this PR.

PR #8113, opened 2026-06-16
Add FuseSplitAttentionToSDPA pass for Intel NPU

enlenama:
Fuse Gemma-style split-attention (separate QK against KV cache and
  the new step's KV, merged via Concat) into a single
  ov::op::v13::ScaledDotProductAttention to enable the NPU NCE path.

  Disabled by default; enabled via the "fuse_split_attention_to_sdpa"
  custom config key on IntelOpenVinoOptions.

  Add test coverage for the pass.

### DIFF
--- a/litert/vendors/intel_openvino/compiler/npu_optimizer.cc
+++ b/litert/vendors/intel_openvino/compiler/npu_optimizer.cc
@@ -14,17 +14,32 @@
 
 #include "litert/vendors/intel_openvino/compiler/npu_optimizer.h"
 
+#include <cstdint>
 #include <memory>
+#include <string>
+#include <vector>
 
+#include "litert/c/internal/litert_logging.h"
 #include "openvino/core/graph_util.hpp"
 #include "openvino/core/model.hpp"
 #include "openvino/core/node.hpp"
 #include "openvino/core/rt_info.hpp"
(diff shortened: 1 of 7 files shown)


litert/vendors/intel_openvino/compiler/npu_optimizer.cc
```diff
@@ -95,11 +122,294 @@ CastIntegerSignToFloat::CastIntegerSignToFloat() {
...
+        add_node->input_value(0).get_node_shared_ptr());
+    auto v_matmul_new = std::dynamic_pointer_cast<ov::op::v0::MatMul>(
+        add_node->input_value(1).get_node_shared_ptr());
+    if (!v_matmul_cache || !v_matmul_new) {
+      return false;
+    }
+
+    // Both V matmuls' first input must come from the same scores splitter.
+    // Two equivalent shapes are accepted:
+    //   (a) a single VariadicSplit producing outputs 0 (cache) and 1 (new);
```
taste1981: VariadicSplit will only exist in internal OV IR, so you should not search for it. Make sure this is removed in the comments and below pattern match code.

taste1981:
@MatthiasKreileder PTAL, thanks!

MatthiasKreileder (approved):

PR #8212, opened 2026-06-21
Make LiteRtLayout ABI identical across MSVC and GCC/Clang (#7459)

### LINKED ISSUE #7459: LiteRtLayout struct has different binary layout under MSVC vs GCC/Clang (non-portable bit-field packing)
## Summary

The public `LiteRtLayout` struct declared in `litert/c/litert_layout.h` is documented as non-opaque ("clients are expected to access struct members directly") but produces **different binary layouts under MSVC vs GCC/Clang**, making the public header binary-incompatible across compilers on the same platform. This affects any consumer that opens a Windows `LiteRt.dll` (MSVC-built) with code compiled by a different compiler (e.g. Dart FFI, Rust bindings, custom MinGW builds).

## Reproduction

```c
typedef struct {
  unsigned int rank : 7;
  bool has_strides : 1;
  int32_t dimensions[LITERT_TENSOR_MAX_RANK];
  uint32_t strides[LITERT_TENSOR_MAX_RANK];
} LiteRtLayout;
```

| Compiler | `offsetof(LiteRtLayout, dimensions)` | `sizeof(LiteRtLayout)` |
|---|---|---|
| Clang (macOS arm64, iOS, Android) | 4 | 68 bytes |
| GCC (Linux x86_64) | 4 | 68 bytes |
| **MSVC (Windows x86_64)** | **8** | **72 bytes** |

The same applies to `LiteRtRankedTensorType`, which embeds `LiteRtLayout`.

## Root cause

MSVC's C bit-field packing rule (see [docs.microsoft.com/cpp/c-language/c-bit-fields](https://learn.microsoft.com/en-us/cpp/c-language/c-bit-fields?view=msvc-170) and [Bruce Dawson's write-up](https://randomascii.wordpress.com/2010/06/06/bit-field-packing-with-visual-c/)) **does not coalesce adjacent bit-fields with different underlying types into the same storage unit**. The `bool has_strides : 1` declared right after `unsigned int rank : 7` therefore opens a fresh 4-byte storage unit on MSVC, while GCC/Clang pack both into the same `unsigned int`.

This is impl-defined behaviour, not a compiler bug.

## Impact

Anyone reading `LiteRtLayout` written by a MSVC-compiled DLL using a non-MSVC ABI (or vice versa) gets garbage:

- `dimensions[0]` looks at the padding of `has_strides`'s storage unit (= 0 or 1) instead of the real first dimension.
- `dimensions[1]` looks at the real `dimensions[0]` (usually = 1, the batch).
- All later dimensions and the entire `strides[]` array are shifted by 4 bytes.

Reported by an embedding-model auto-detection path that returned `seqLen=1, dim=1` on Windows for a `.tflite` model that was clearly `[1, 256]` × `[1, 768]` (Gecko / EmbeddingGemma). The same Dart FFI code produced correct dimensions on macOS / Linux / iOS / Android because all of those use GCC/Clang.

## Suggested fixes (any one would resolve it)

1. **Change the storage type of `has_strides` to match `rank`** — pure source change, no ABI break on GCC/Clang:
   ```c
   unsigned int rank : 7;
   unsigned int has_strides : 1;
   ```
   MSVC will pack both into one `unsigned int` and `dimensions[]` lands at offset 4 everywhere.

2. **Expose accessor functions** in the public C API:
   ```c
   LITERT_API LiteRtStatus LiteRtLayoutGetRank(const LiteRtLayout*, uint8_t* rank);
   LITERT_API LiteRtStatus LiteRtLayoutGetDimension(const LiteRtLayout*, int32_t index, int32_t* dim);
   LITERT_API LiteRtStatus LiteRtLayoutGetHasStrides(const LiteRtLayout*, bool* has_strides);
   LITERT_API LiteRtStatus LiteRtLayoutGetStride(const LiteRtLayout*, int32_t index, uint32_t* stride);
   ```
   The header note ("access methods do not need to and should NOT be exported") can be revisited — non-opaque struct + non-portable ABI is a hazard, not a feature.

3. **Document the cross-compiler ABI hazard** explicitly in the header and recommend wrapping in compiler-specific structs.

(1) is the cheapest and least disruptive.

## Workaround on consumer side (for reference)

`flutter_gemma` 0.15.2 ([release notes](https://github.com/DenisovAV/flutter_gemma/releases/tag/v0.15.2)) ships dual Dart FFI struct definitions and picks the right one at allocation time via `Platform.isWindows`. Source: [lib/core/litert/litert_bindings.dart](https://github.com/DenisovAV/flutter_gemma/blob/v0.15.2/lib/core/litert/litert_bindings.dart). This works but doubles the type surface and isn't something every consumer should have to write.

## Refs

- [LiteRT header](https://github.com/google-ai-edge/LiteRT/blob/main/litert/c/litert_layout.h)
- [MSVC bit-field packing docs](https://learn.microsoft.com/en-us/cpp/c-language/c-bit-fields?view=msvc-170)
- [Bruce Dawson — Bit-field Packing with Visual C++](https://randomascii.wordpress.com/2010/06/06/bit-field-packing-with-visual-c/)
- Chromium discussion of the same hazard: https://groups.google.com/a/chromium.org/g/style-dev/c/TB0Uva6pAus

ArsalanShakil:
## What

Fixes #7459.

The public `LiteRtLayout` struct in `litert/c/litert_layout.h` declares:

```c
unsigned int rank : 7;
bool         has_strides : 1;
```

MSVC does not coalesce adjacent bit-fields with **different underlying types** into the same storage unit. So on MSVC the `bool` field opens a fresh 4-byte storage unit, making the struct **72 bytes with `dimensions` at offset 8**, versus **68 bytes / offset 4** on GCC/Clang. `LiteRtRankedTensorType` (which embeds `LiteRtLayout`) diverges the same way: 76 vs 72 bytes.

| Compiler | `sizeof(LiteRtLayout)` | `offsetof(dimensions)` |
|---|---|---|
| Clang / GCC | 68 | 4 |
| MSVC | 72 | 8 |

Because this is a documented **non-opaque, ABI-stable public type**, any consumer that reads a struct produced by a differently-compiled `LiteRt.dll` (Dart FFI, Rust bindings, MinGW builds) sees every dimension shifted by 4 bytes. The current header (commit `edc801ec1`) only *documents* the divergence via per-compiler `static_assert` branches; it does not fix the incompatibility.

## Change

Give `has_strides` the same underlying type as `rank` so both pack into a single storage unit on every compiler:

```c
unsigned int rank : 7;
unsigned int has_strides : 1;
```

- **No-op for the GCC/Clang ABI** — still 68 bytes, `dimensions` at offset 4.
- **Unifies MSVC** to the same 68-byte layout.
- Collapses the per-compiler `static_assert` branches in `litert_layout.h` and `litert_model_types.h` into a single expectation, so any future reintroduction of cross-compiler divergence fails the build.

This is the minimal, lowest-risk option from the issue's suggested fixes (no public API surface added, no behavior change). Assigning `bool` values to / reading the field as `bool` continues to work (the `litert/cc` wrapper is unaffected).

## Verification

Compiled the real headers under clang with the asserts active:

```
LiteRtLayout sizeof=68 off(dim)=4 off(str)=36
LiteRtRankedTensorType sizeof=72 off(layout)=4
```

Confirmed identical layout before and after the change on clang, and that a `true` assigned to the 1-bit `unsigned int` field reads back as `1`.

Note: I don't have an MSVC environment to reproduce the 72→68 unification directly; that follows from the documented MSVC bit-field packing rule and the measurements in #7459. The in-header `static_assert`s serve as the compile-time regression guard across CI toolchains.

### DIFF
--- a/litert/c/litert_layout.h
+++ b/litert/c/litert_layout.h
@@ -33,7 +33,16 @@ extern "C" {
 /// @note This concrete type is part of the public API and is ABI stable.
 typedef struct {
   unsigned int rank : 7;  // The number of dimensions.
-  bool has_strides : 1;   // Whether the layout has strides.
+  // Whether the layout has strides.
+  //
+  // NOTE: This bit-field uses `unsigned int` (matching `rank` above) rather
+  // than `bool` on purpose. MSVC does not coalesce adjacent bit-fields with
+  // different underlying types into the same storage unit, so a `bool` field
+  // here would open a fresh 4-byte unit and shift `dimensions`/`strides`,
+  // making this public struct binary-incompatible between MSVC and GCC/Clang
+  // builds. Keeping the underlying type identical packs both fields together
+  // on every compiler. See https://github.com/google-ai-edge/LiteRT/issues/7459
+  unsigned int has_strides : 1;
 
   // Dimension sizes, array of length `rank`. Dynamic dimensions are anything
   // less than 0. Everything from [rank, LITERT_MAX_RANK) is undefined.
@@ -43,21 +52,18 @@ typedef struct {
   uint32_t strides[LITERT_TENSOR_MAX_RANK];
 } LiteRtLayout;
 
+// The layout below is now identical across MSVC and GCC/Clang because `rank`
+// and `has_strides` share an underlying type and pack into a single storage
+// unit on every compiler. These asserts intentionally have no `_MSC_VER`
+// branch so any future change that reintroduces cross-compiler divergence
+// fails the build.
 #if defined(__cplusplus) && defined(__SIZEOF_POINTER__) && \
     __SIZEOF_POINTER__ == 8
-#if !defined(_MSC_VER)
 static_assert(sizeof(LiteRtLayout) == 68, "LiteRtLayout size mismatch");
 static_assert(offsetof(LiteRtLayout, dimensions) == 4,
               "LiteRtLayout dimensions offset mismatch");
 static_assert(offsetof(LiteRtLayout, strides) == 36,
               "LiteRtLayout strides offset mismatch");
-#else   // !defined(_MSC_VER)
-static_assert(sizeof(LiteRtLayout) == 72, "LiteRtLayout size mismatch");
-static_assert(offsetof(LiteRtLayout, dimensions) == 8,
-              "LiteRtLayout dimensions offset mismatch");
-static_assert(offsetof(LiteRtLayout, strides) == 40,
-              "LiteRtLayout strides offset mismatch");
-#endif  // !defined(_MSC_VER)
 #endif  // __cplusplus
 
 // Return the number of scalar elements in the provided tensor layout. Return an
--- a/litert/c/litert_model_types.h
+++ b/litert/c/litert_model_types.h
@@ -81,15 +81,14 @@ typedef struct {
   LiteRtLayout layout;
 } LiteRtRankedTensorType;
 
+// `LiteRtRankedTensorType` embeds `LiteRtLayout`, whose layout is now identical
+// across MSVC and GCC/Clang (see litert/c/litert_layout.h), so this size no
+// longer diverges by compiler. The assert intentionally has no `_MSC_VER`
+// branch.
 #if defined(__cplusplus) && defined(__SIZEOF_POINTER__) && \
     __SIZEOF_POINTER__ == 8
-#if !defined(_MSC_VER)
 static_assert(sizeof(LiteRtRankedTensorType) == 72,
               "LiteRtRankedTensorType size mismatch");
-#else   // !defined(_MSC_VER)
-static_assert(sizeof(LiteRtRankedTensorType) == 76,
-              "LiteRtRankedTensorType size mismatch");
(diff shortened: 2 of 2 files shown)


snnn (approved):

ArsalanShakil:
Will solve the failing runs soon, away on vacation.

ArsalanShakil:
@snnn solved the failing runs

snnn:
Is this change already merged into the main branch or not?

ArsalanShakil:
@snnn No, it is not merged into `main` yet, the change currently lives only on this PR branch. `litert/c/litert_layout.h` on `main` still has the old bitfield layout (`unsigned int rank : 7; bool has_strides : 1;`).

The PR is approved and CI is green (the stale clang-tidy failure was from an earlier run; the latest checks pass). It also imported cleanly via Copybara (cl/951886377).
@snnn please merge this PR as i don't have permissions to do so.

rwitcher:
Hi @ArsalanShakil - Thank you for this PR. It was actually merged via a batch PR #8582, so this change is merged and synced with main. Closing this but feel free to reach out with any other questions.

PR #8258, opened 2026-06-23
Fix parsing of Intel OpenVINO configs_map entries

enlenama:
absl::StrSplit(pair, '=') split each comma-separated entry on every '=', so any pair
whose value itself contained '=' produced more than two tokens and was rejected
and dropped with a warning. This affects any property whose value embeds further
 '=' characters, e.g. nested key=value sub-lists like
NPU_COMPILATION_MODE_PARAMS=enable-flash-sdpa-conversion=true.

Split on the first '=' only so the leading delimiter separates the key from the value
and the value passes through intact. Updates the malformed-pair test for the new
splitting behavior and adds a regression test. 

Also runs clang-format (Google style) over the touched files.

### DIFF
--- a/litert/tools/flags/vendors/intel_openvino_flags.cc
+++ b/litert/tools/flags/vendors/intel_openvino_flags.cc
@@ -145,9 +145,13 @@ Expected<void> UpdateIntelOpenVinoOptionsFromFlags(
     std::vector<std::string> config_pairs =
         absl::StrSplit(configs_map_str, ',');
     for (const auto& pair : config_pairs) {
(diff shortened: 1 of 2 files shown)


enlenama:
Hi @MatthiasKreileder, this PR fixes the OpenVINO config parsing issue. PTAL, thanks!

MatthiasKreileder (approved):
Thanks for the fix!

MatthiasKreileder:
Merged via https://github.com/google-ai-edge/LiteRT/pull/8582/changes#diff-61bad46a0636082bb6ee363e68b66a2b71c4d4e81039086dd4a1d685e071c645 - thanks for the fix

PR #8260, opened 2026-06-23
[Intel OV] Add support for SHLOComposite op

akodanka:
STABLEHLO_COMPOSITE odml.rms_norm translation has now been added to Openvino :
https://github.com/openvinotoolkit/openvino/pull/36263

Enabling this op from the LiteRT intel_openvino compiler side.

Validations done :
1. Used LiteRT-LM benchmark tool for capturing the performance on Gemma4 E2B model. Observed similar results, with slight improvement in decode speed and memory usage, and slight drop in the prefill speed.
2. Also verified that sample prompts are giving the same result as before.


MatthiasKreileder (commented):
Thank you for the effort, this look good.

Given that this will change how our 'launched' LLMs like Gemma4 2B are lowered from LiteRT to OpenVino, it would best to smoke test it once.

Could you provide a brief section in the PR description on the verification for this change?

MatthiasKreileder (approved):

litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -97,19 +97,22 @@ litert_dynamic_lib(
...
         "//litert/c:litert_op_code",
+        "//litert/c:litert_op_options_header",
         "//litert/c/internal:litert_logging",
         "//litert/c/internal:litert_logging_helper_with_compiler_context",
         "//litert/c/options:litert_intel_openvino_options",
         "//litert/cc:litert_expected",
         "//litert/cc:litert_macros",
         "//litert/cc/internal:litert_context_wrapper",
         "//litert/cc/internal:litert_handle",
+        "//litert/cc/internal:litert_op_options",
```
MatthiasKreileder, line 108: Could we try without adding new deps on internal targets?

If this is for the rms norm op name you can do like this https://github.com/google-ai-edge/LiteRT/blob/51d574b47178d1a0395e842150e5c8077459aa23/litert/vendors/mediatek/compiler/create_model.cc#L312

litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -97,19 +97,22 @@ litert_dynamic_lib(
         "//litert/c:litert_builder_header",
         "//litert/c:litert_common",
         "//litert/c:litert_op_code",
+        "//litert/c:litert_op_options_header",
```
MatthiasKreileder, line 100: Don't we need to add the same deps to `LiteRtCompilerPlugin` than we do here?

MatthiasKreileder (changes requested):
A few nits

litert/vendors/intel_openvino/compiler/BUILD
akodanka: Looks like it was still able to access the header through the graph_iterator->decoder dependency.
I'll still add that now.

litert/vendors/intel_openvino/compiler/BUILD
akodanka: I was trying to avoid that hardcoding, and referred the code from here 
https://github.com/google-ai-edge/LiteRT/blob/3d9c400ed4c74a676bc0ccf9a63aa81c2c6ae7a9/litert/vendors/qualcomm/compiler/qnn_compose_graph.cc#L875

Do let me know if you still suggest using the direct string.

PR #8330, opened 2026-06-25
[OV] Support TopK-v2 with constant K.

taste1981:
Intel NPU supports TopK-v2 when the K is a constant input. Rejects when K is a dynamic input.


litert/vendors/intel_openvino/compiler/openvino_compiler_plugin_test.cc
```diff
@@ -88,6 +89,62 @@ TEST(TestOVPlugin, CompileAddSubgraph) {
   LiteRtDestroyCompiledResult(compiled);
 }
 
+// TOPK_V2 with a constant K (input 1 carries a weights buffer) yields a static
+// output shape and must be selected for the OpenVINO partition.
+TEST(TestOVPlugin, PartitionTopKConstK) {
+  auto plugin = CreatePlugin(LrtGetCompilerContext());
+  auto model = testing::LoadTestFileModel("simple_topk_op.tflite");
```
MatthiasKreileder, line 96: Thanks for creating a standalone test.

Would it be difficult to create an MLIR file instead that gets turned into a tflite by blaze? e.g.:
- https://github.com/google-ai-edge/LiteRT/blob/main/litert/test/testdata/one_mul.mlir
- https://github.com/google-ai-edge/LiteRT/blob/3c1752349c6f2b73e257989d048b1e7738df9722/litert/vendors/examples/end_to_end_test.cc#L36

If this is tedious to do I am happy to stick with the tflite you added.

MatthiasKreileder (commented):
Thanks! Looks good. Left one comment w.r.t testing.

Also please rebase to resolve the decoder conflict.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin_test.cc
taste1981: Fixed the decoder conflict.  

For the .mlir, I added the file side-by-side with the .tflite (similar as the simple_topk_op.tflite |.mlir which already exists prior to my change). 
However, it seems only copybara will really generate the model from .mlir to .tflite... on Windows it seems tf_tfl_translate transitively depends on net_zstd, whose huf_decompress.S is fed to MASM and fails to assemble.

I would like this PR just stop here, and handle this as a follow up. WDYT?

MatthiasKreileder: > handle this as a follow up

Thanks for trying, then don't bother. I was hoping it's already working e2e, it's not critical.

MatthiasKreileder (approved):

PR #8331, opened 2026-06-25
Add OptimizeMHAGemma4BPrefill by removing Quantize Op

qti-hungjuiw:


qti-hungjuiw:
Close since we will use https://github.com/google-ai-edge/LiteRT/pull/8653

PR #8389, opened 2026-06-29
Update Image Preprocessor

Jiseong-oh:
- Add an optional NormalizationConfig (rescale factor + per-channel mean/std) to the image preprocessor so encoders with custom normalization (e.g. LFM2 VL) are supported.
- Make the patchify positions_xy tensor optional via an emit_positions flag, since single-input encoders consume only the images tensor while ViT encoders also need positions.
- Unify the resize/rescale/normalize path in the STB preprocessor and branch on target dimensions instead of patchify config, keeping existing ViT and fixed-size models unchanged.

### DIFF
--- a/support/preprocessor/image_preprocessor.h
+++ b/support/preprocessor/image_preprocessor.h
@@ -18,6 +18,7 @@
 #include <memory>
 #include <optional>
 #include <utility>
+#include <vector>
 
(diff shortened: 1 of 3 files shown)


yichunk (commented):
Thanks for the update!

Could you add unit tests for the new update?
testing the optional position tensors, and the normalization config.

PR #8613, opened 2026-07-10
LSTM: check input-to-output weights rank before reading dims->data[0]

### LINKED ISSUE #8610: LSTM / UNIDIRECTIONAL_SEQUENCE_LSTM: heap OOB read of weights rank in Prepare
Both LSTM `Prepare` functions read the input-to-output weights leading dimension before checking the tensor's rank:

- `tflite/kernels/lstm.cc:1329` -> `n_cell = input_to_output_weights->dims->data[0];`
- `tflite/kernels/unidirectional_sequence_lstm.cc:935` -> same

In both, the `TF_LITE_ENSURE_EQ(dims->size, 2)` is on the next line, one statement too late. A weights tensor of rank 0 reads past the `dims` array.

Repro: a model with an LSTM (or UNIDIRECTIONAL_SEQUENCE_LSTM) whose input-to-output weights tensor is rank 0. On an ASan build, `AllocateTensors()` reports a heap-buffer-overflow READ in the respective `Prepare` (LiteRT HEAD d419a5d, x86-64).

Suggested fix: move the `dims->size == 2` check above the `dims->data[0]` read in both files. PR to follow.

professor-moody:
`lstm.cc` and `unidirectional_sequence_lstm.cc` read `input_to_output_weights->dims->data[0]` before the `dims->size == 2` check on the next line, so a rank-0 weights tensor reads past the `dims` array. This moves the size check above the read in both files.

Verified on an ASan build: both repros (a rank-0 input-to-output weights tensor) now return `kTfLiteError` cleanly.

Fixes #8610

### DIFF
--- a/tflite/kernels/lstm.cc
+++ b/tflite/kernels/lstm.cc
@@ -1326,8 +1326,8 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
   TF_LITE_ENSURE_OK(context,
                     GetInputSafe(context, node, kInputToOutputWeightsTensor,
                                  &input_to_output_weights));
-  const int n_cell = input_to_output_weights->dims->data[0];
   TF_LITE_ENSURE_EQ(context, input_to_output_weights->dims->size, 2);
+  const int n_cell = input_to_output_weights->dims->data[0];
   TF_LITE_ENSURE_EQ(context, input_to_output_weights->dims->data[1], n_input);
 
   const TfLiteTensor* recurrent_to_output_weights;
--- a/tflite/kernels/unidirectional_sequence_lstm.cc
+++ b/tflite/kernels/unidirectional_sequence_lstm.cc
@@ -932,8 +932,8 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
       context,
       GetInputSafe(context, node, lstm::full::kInputToOutputWeightsTensor,
                    &input_to_output_weights));
-  const int n_cell = input_to_output_weights->dims->data[0];
   TF_LITE_ENSURE_EQ(context, input_to_output_weights->dims->size, 2);
+  const int n_cell = input_to_output_weights->dims->data[0];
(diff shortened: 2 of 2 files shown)


professor-moody:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google-ai-edge/LiteRT/pull/8613/checks?check_run_id=86392150981) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

@googlebot I signed it!

professor-moody:
Adding a note on scope: there is a third file with the same pattern that this PR does not touch, and I should have said so up front rather than leaving it implicit.

This PR fixes `lstm.cc` and `unidirectional_sequence_lstm.cc`, where the rank of `input_to_output_weights` is read before it is validated:

```cpp
const int n_cell = input_to_output_weights->dims->data[0];        // read
TF_LITE_ENSURE_EQ(context, input_to_output_weights->dims->size, 2);  // checked after
```

`bidirectional_sequence_lstm.cc` has the identical ordering at two sites, one per direction:

```
:444  const int n_fw_cell = fw_input_to_output_weights->dims->data[0];
:445  TF_LITE_ENSURE_EQ(context, fw_input_to_output_weights->dims->size, 2);

:453  const int n_bw_cell = bw_input_to_output_weights->dims->data[0];
:454  TF_LITE_ENSURE_EQ(context, bw_input_to_output_weights->dims->size, 2);
```

Worth noting that the same file gets it right elsewhere: at `:230` the rank check precedes the `dims->data[0]` read for the same tensor. So this is an ordering slip at two sites rather than a missing convention.

I have not built or exercised the bidirectional path, so I am reporting the pattern rather than claiming a witnessed crash there. Happy to extend this PR to cover those two sites, or to leave them for a separate change if you would rather keep this one narrow. Let me know which you prefer.

PR #8651, opened 2026-07-13
[OpenVino] Convert i2 weights to i4 for OpenVINO GPU

zlma7001:
OpenVINO GPU does not support i2 or u2 weights directly, so convert them to i4 to make the generated model more compatible with the OV GPU backend.

Test commands: 
.\apply_plugin_main.exe --cmd apply --model=.\models\model.tflite       -o .\intel_gemma4_gpu_aot.tflite --libs .\ --intel_openvino_graph_backends="gpu" --soc_manufacturer "IntelOpenVINO"
.\litertlm_builder_cli.exe ... output --path .\intel_gemma4_gpu_aot.litertlm
.\litert_lm_advanced_main.exe --backend=npu --model_path=.\intel_gemma4_gpu_aot.litertlm --use_hw_masking_for_npu=false --benchmark --benchmark_prefill_tokens=3840 --benchmark_decode_tokens=256

### DIFF
--- a/litert/vendors/intel_openvino/compiler/graph_iterator.h
+++ b/litert/vendors/intel_openvino/compiler/graph_iterator.h
@@ -19,6 +19,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <string>
 #include <vector>
 
 #include "openvino/frontend/tensorflow_lite/decoder.hpp"
@@ -47,8 +48,9 @@ class GraphIteratorDelegate
     : public ov::frontend::tensorflow_lite::GraphIterator {
  public:
   GraphIteratorDelegate(const LiteRtCompilerContext* ctx,
-                        const litert::compiler::Subgraph* graph)
-      : ctx_(ctx), subgraph_ptr_(graph) {
+                        const litert::compiler::Subgraph* graph,
+                        std::string device = "NPU")
+      : ctx_(ctx), subgraph_ptr_(graph), device_(std::move(device)) {
     for (const auto& input : subgraph_ptr_->Inputs()) {
       if (input.IsSubgraphInput()) {
         iterator_indices_.input_index_++;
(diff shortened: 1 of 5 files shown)


litert/vendors/intel_openvino/compiler/graph_iterator.h
```diff
@@ -96,13 +98,32 @@ class GraphIteratorDelegate
   };
 
  private:
+  // Rewrites signed i2 weights to unsigned u2 (NPU/CPU path): flips the MSB
+  // of every 2-bit element (XOR 0xAA) and shifts the zero points by +2 so
+  // the dequantized values are unchanged.  Updates |tensor_meta_info| in
+  // place to point at the converted buffer with element type u2.
+  void ConvertI2WeightsToU2(
```
MatthiasKreileder, line 105: Is there any way to add a test driver for the graph_iterator? Coverage is quite low which can cause problems the more logic we add in here.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -521,7 +521,8 @@ LiteRtStatus LiteRtCompilerPluginCompile(
         std::shared_ptr<ov::frontend::tensorflow_lite::GraphIterator>
             graph_delegate =
                 std::make_shared<litert::openvino::GraphIteratorDelegate>(
-                    compiler_plugin->ctx(), &expected_subgraph.Value());
+                    compiler_plugin->ctx(), &expected_subgraph.Value(),
```
MatthiasKreileder, line 524: Any way to add an automatic test for compilation? IIUC it is possible to run the compiler plugin on a host with an NPU (which is the case for the github action hosts)?

If this is unreasonable we don't need to block on this, but coverage is something we need to improve.

MatthiasKreileder (changes requested):
Thanks for the work, the change looks good.

Could you please add a manual test plan to the PR description? I.e. the command to repro / trigger these code paths.

litert/vendors/intel_openvino/compiler/graph_iterator.h
zlma7001: Added a parameterized graph_iterator_test using the existing FFW-2-bit.tflite fixture and manually test passed. It exercises both GPU (i2 to i4) and NPU (i2 to u2) paths and validates the converted bytes, element types, and zero points.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
zlma7001: Currently OV doesn't support off-host GPU compilation, so this can be only test manually.

zlma7001:
Thanks @MatthiasKreileder , test commands updated in message, PTAL

MatthiasKreileder (approved):

PR #8745, opened 2026-07-16
Weight sharing ovgpu

anikulk:
Enable weight sharing leveraging Global Graphs and Intel GPU  ov::intel_gpu::SharedMemType::USM_HOST_BUFFER.

### DIFF
--- a/litert/vendors/intel_openvino/compiler/BUILD
+++ b/litert/vendors/intel_openvino/compiler/BUILD
@@ -102,6 +102,18 @@ litert_test(
 )
 
 litert_lib(
+    name = "weights_to_parameters",
+    srcs = ["weights_to_parameters.cc"],
+    hdrs = ["weights_to_parameters.h"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    deps = [
+        ":weight_bank",
+        "@intel_openvino//:openvino",
+    ],
+)
+
+litert_lib(
     name = "openvino_compile_context",
     srcs = ["openvino_compile_context.cc"],
     hdrs = ["openvino_compile_context.h"],
--- a/litert/vendors/intel_openvino/compiler/BUILD
+++ b/litert/vendors/intel_openvino/compiler/BUILD
@@ -64,6 +64,44 @@ litert_test(
 )
 
 litert_lib(
+    name = "weight_bank",
+    srcs = ["weight_bank.cc"],
+    hdrs = ["weight_bank.h"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    deps = [
+        "//litert/compiler/cc:litert_model",
+        "@com_google_absl//absl/types:span",
+    ],
+)
+
+litert_test(
+    name = "weight_bank_test",
+    srcs = ["weight_bank_test.cc"],
+    copts = ["-fexceptions"],
+    data = [
+        "//litert/test:mlir_test_data",
+        "//litert/test:tflite_test_data",
+    ],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    tags = [
+        "manual",
+        "nobuilder",
+        "notap",
+    ],
+    deps = [
+        ":weight_bank",
+        "//litert/c:litert_runtime_c_api_shared_lib",
+        "//litert/c/internal:litert_compiler_context",
+        "//litert/cc/dynamic_runtime:litert_extended_model",
+        "//litert/compiler/cc:litert_model",
+        "//litert/test:common",
+        "//litert/test:load_test_model",
+        "//litert/test:test_models",
+    ],
+)
+
+litert_lib(
     name = "openvino_compile_context",
     srcs = ["openvino_compile_context.cc"],
     hdrs = ["openvino_compile_context.h"],
--- a/litert/vendors/intel_openvino/compiler/BUILD
+++ b/litert/vendors/intel_openvino/compiler/BUILD
@@ -102,6 +102,33 @@ litert_test(
 )
 
 litert_lib(
+    name = "global_graph",
+    srcs = ["global_graph.cc"],
+    hdrs = ["global_graph.h"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    visibility = ["//litert/vendors/intel_openvino:__subpackages__"],
+    deps = [
+        "//litert/cc:litert_expected",
+    ],
+)
+
+litert_test(
+    name = "global_graph_test",
+    srcs = ["global_graph_test.cc"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    tags = [
+        "manual",
+        "nobuilder",
+        "notap",
+    ],
+    deps = [
+        ":global_graph",
+    ],
+)
+
+litert_lib(
     name = "weights_to_parameters",
     srcs = ["weights_to_parameters.cc"],
     hdrs = ["weights_to_parameters.h"],
@@ -109,6 +136,35 @@ litert_lib(
     features = ["-use_header_modules"],  # Incompatible with -fexceptions.
     deps = [
         ":weight_bank",
+        "@intel_openvino//:openvino",
+    ],
+)
+
+litert_test(
+    name = "weights_to_parameters_test",
+    srcs = ["weights_to_parameters_test.cc"],
+    copts = ["-fexceptions"],
+    data = [
+        "//litert/test:mlir_test_data",
+        "//litert/test:tflite_test_data",
+    ],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    tags = [
+        "manual",
+        "nobuilder",
+        "notap",
+    ],
+    deps = [
+        ":graph_iterator",
+        ":weight_bank",
+        ":weights_to_parameters",
+        "//litert/c:litert_runtime_c_api_shared_lib",
+        "//litert/c/internal:litert_compiler_context",
+        "//litert/cc/dynamic_runtime:litert_extended_model",
+        "//litert/compiler/cc:litert_model",
+        "//litert/test:common",
+        "//litert/test:load_test_model",
+        "//litert/test:test_models",
         "@intel_openvino//:openvino",
     ],
 )
@@ -160,6 +216,9 @@ litert_dynamic_lib(
     deps = [
         ":graph_iterator",
         ":openvino_compile_context",
+        ":weight_bank",
+        ":weights_to_parameters",
+        ":global_graph",
         ":openvino_soc_config",
         "//litert/c:litert_builder_header",
         "//litert/c:litert_common",
@@ -206,6 +265,9 @@ litert_bin(
         ":litert_compiler_plugin_c_api",
         ":openvino_compile_context",
         ":openvino_soc_config",
+        ":weight_bank",
+        ":weights_to_parameters",
+        ":global_graph",
         "//litert/c:litert_common",
         "//litert/c:litert_op_code",
         "//litert/c:litert_op_options_header",
--- a/litert/vendors/intel_openvino/compiler/BUILD
+++ b/litert/vendors/intel_openvino/compiler/BUILD
@@ -98,6 +98,7 @@ litert_test(
         "//litert/test:common",
         "//litert/test:load_test_model",
         "//litert/test:test_models",
+        "@intel_openvino//:openvino",
     ],
 )
 
@@ -118,11 +119,9 @@ litert_test(
     srcs = ["global_graph_test.cc"],
     copts = ["-fexceptions"],
     features = ["-use_header_modules"],  # Incompatible with -fexceptions.
-    tags = [
-        "manual",
-        "nobuilder",
-        "notap",
-    ],
+    # No OpenVINO runtime dependency (pure container serialize/parse logic), so
+    # this builds and runs in CI on plain CPU -- unlike the other intel_openvino
+    # tests, which link @intel_openvino and need the OV runtime/hardware.
     deps = [
         ":global_graph",
     ],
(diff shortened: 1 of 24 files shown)


litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
```diff
@@ -0,0 +1,121 @@
...
+
+litert::Expected<std::vector<BoundWeight>> BindSharedWeightsGpu(
+    ov::Core& core, const OpenVinoGlobalGraph& global_graph,
+    const ov::CompiledModel& compiled_model,
+    const std::map<uint32_t, uint32_t>& const_map) {
+  ov::RemoteContext ctx = core.get_default_context("GPU");
+
+  {
+    absl::MutexLock lock(&GpuBankMutex());
+    GpuBank& bank = SharedGpuBank();
```
taste1981: This may finally be used by Chromium in a single process. Since SharedGpuBank() is a single process-global bank gated by bank.ready, keyed by nothing that identifies the container, if two different models load in the same process, the second sees ready == true and silently reuse first model's USM contents and offsets.
Make sure you key the bank by container identity (e.g., a hash of the buffer pool or something else) or scope it to the device/invocation context rather than the process.

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
```diff
@@ -0,0 +1,121 @@
...
+      }
+      bank.ready = true;
+      LITERT_LOG(LITERT_INFO,
+                 "GlobalGraph: allocated shared usm-host bank (%zu bytes)",
+                 total);
+    }
+  }
+
+  GpuBank& bank = SharedGpuBank();
+  auto* base = static_cast<uint8_t*>(bank.base);
```
taste1981: Why `bank.base` & `bank.offset` is allowed to be accessed out of the Mutex protection? This violates the ABSL_EXCLUSIVE_LOCKS_REQUIRED(GpuBackMutex()) annotation.

litert/vendors/intel_openvino/dispatch/invocation_context.cc
```diff
@@ -139,6 +146,49 @@ LiteRtDispatchInvocationContextT::Create(
...
+                          static_cast<const uint8_t*>(exec_bytecode_ptr),
+                          exec_bytecode_size));
+
+    const litert::openvino::OpenVinoGlobalGraph::Subgraph* selected = nullptr;
+    if (function_name != nullptr && function_name[0] != '\0') {
+      auto it = global_graph->subgraphs.find(function_name);
+      if (it != global_graph->subgraphs.end()) selected = &it->second;
+    }
+    if (selected == nullptr && !global_graph->subgraphs.empty()) {
+      selected = &global_graph->subgraphs.begin()->second;  // fall back to first
```
taste1981: For a genuine muilti-partition model, won't a null/mismatched `function_name` makes every partition load `Partition_0`'s payload/const_map/device? Maybe you should only fallback when `subgraphs.size() == 1`; otherwise return an error?

litert/vendors/intel_openvino/compiler/weight_bank.cc
```diff
@@ -0,0 +1,64 @@
...
+    for (const auto& input : op.Inputs()) {
+      if (!input.HasWeights()) {
+        continue;
+      }
+      const auto weights = input.Weights();
+      // Keyed by BufferId, so a buffer shared by multiple ops/partitions is
+      // recorded once. The bytes are identical for a given id, so re-assignment
+      // is harmless.
+      const int32_t buffer_id = weights.BufferId();
+      buffer_bytes_[buffer_id] = weights.Bytes();
```
taste1981, line 44: Weights::BufferId() returns -1 as its default/failure sentinel sometimes so a guard against buffer_id <0 better be added here.

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
```diff
@@ -0,0 +1,121 @@
...
+  std::vector<BoundWeight> bound;
+  const auto& inputs = compiled_model.inputs();
```
taste1981, line 76: `const_map` keys are positions in `model->inputs()` at compile time; at dispatch they index `compiled_model.inputs()`. It looks like you rely on `core.import_model` preserving parameter order exactly? If that's guranteed - fine; but if not, you'll need to consider kerying `const_map` by the parameter's friendly_name instead of positional index (which is more robust).

litert/vendors/intel_openvino/compiler/global_graph.h
```diff
@@ -0,0 +1,77 @@
...
+
+  // Serialize the whole container to one blob (see layout above).
+  std::string Serialize() const;
+
+  // Parse a container blob. Returns an error if magic/bounds are invalid.
+  static litert::Expected<OpenVinoGlobalGraph> Parse(const uint8_t* data,
+                                                     size_t size);
+
+  // Fast check: does |data| begin with the OVGLOBAL magic?
+  static bool HasMagic(const uint8_t* data, size_t size);
```
taste1981, line 76: You have magic here which is good, but missing a version byte which makes future layout change not possible. Please add a uint16 version as well for future extension and backward compatibility.

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
```diff
@@ -0,0 +1,121 @@
...
+  {
+    absl::MutexLock lock(&GpuBankMutex());
+    GpuBank& bank = SharedGpuBank();
+    if (!bank.ready) {
+      size_t total = 0;
+      for (const auto& [buffer_id, bytes] : global_graph.buffers)
+        total += bytes.size();
+      // Allocate the pool as one usm-host buffer of bytes; per-weight views
+      // reinterpret slices as the weight's element type below.
+      bank.usm = ctx.create_tensor(
```
taste1981: Could you confirm there won't be potential issue for such kind of USM view for each weight? Will an F16/F32 weight starting at an unaligned offset violate GPU driver alignement expectations?
I'm just concerning whether this will cause a potential copy when you create the remote tensor below, if you don't align each pool entry to at least the max element size (or a cache line) ...

litert/vendors/intel_openvino/compiler/weight_bank.h
```diff
@@ -0,0 +1,80 @@
+// Copyright 2026 Google LLC.
```
taste1981, line 1: The license header for files you added uses two different licenses: Google's and Intel's. Could you consolidate to one? Same for other newly added files.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -51,6 +52,9 @@
 #include "litert/vendors/intel_openvino/compiler/graph_iterator.h"
 #include "litert/vendors/intel_openvino/compiler/openvino_compile_context.h"
 #include "litert/vendors/intel_openvino/compiler/openvino_soc_config.h"
+#include "litert/vendors/intel_openvino/compiler/global_graph.h"
```
taste1981: style: include should be arranged in alphabet order.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -504,6 +514,32 @@ LiteRtStatus LiteRtCompilerPluginCompile(
...
+    // Cross-partition weight sharing (opt-in via LITERT_OV_EMBED_WEIGHTS=1):
+    // all partitions' distinct weights are deduplicated by LiteRt BufferId into
+    // one OpenVinoGlobalGraph container (shared buffer pool + per-subgraph
+    // {payload, const_map, device}). The same serialized blob is returned for
+    // every partition; the dispatcher selects its subgraph and resolves the
+    // subgraph's weights against the shared pool.
+    litert::openvino::WeightBank weight_bank;
+    litert::openvino::OpenVinoGlobalGraph global_graph;
+    const char* embed_env = std::getenv("LITERT_OV_EMBED_WEIGHTS");
+    const bool share_weights = embed_env != nullptr && embed_env[0] == '1';
```
taste1981: strcmp(embed_env, "1") reads better? You should not be expecting "10" or "1x" here.

taste1981 (commented):
Weights copied 2–3× in host memory during model compilation?

IIUC, weight bytes (potentially multi-GB) are copied from the mmap into `global_graph.buffers` as std::string, then again into the serialized blob, on top of the still-mapped originals. Given this potential >2GB handling constraints, consider serializing in a single pass (reserve output, append buffer views directly) or streaming into byte_code[0].

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
anikulk: Thanks for this perspective, I made the GpuSharedBank in the scope of the device context. There will be a single bank for each model and thus we will be able to share it. If chromium has single process with multiple thread for different model too, there would be 2 device contexts so the synchronization here should work.

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
anikulk: Done.

litert/vendors/intel_openvino/dispatch/invocation_context.cc
anikulk: Thanks, took care of this.

litert/vendors/intel_openvino/compiler/weight_bank.cc
anikulk: This should not happen, but I added a gaurd.

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
anikulk: Done.

litert/vendors/intel_openvino/compiler/global_graph.h
anikulk: Done

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
anikulk: Can I address this in a future fix? I want to do some more testing. It is not an issue in the INT4 weights. Let me know.

taste1981: Yes sure. Please place a //TODO here so this is recorded.

litert/vendors/intel_openvino/dispatch/weight_bank_runtime.cc
taste1981: My understanding is that, the ov::Core in LiteRT/OV dispatch is singleton per-process. Once you invoke get_default_context("GPU") in two threads, underlyingly the returned remote context is the same.
Please make sure USM tensor caching on OV side won't cause any problem, in case two models coincidentally have tensors of the the same signature/shape.

anikulk: Yes ov::Core is singleton, and ov::Core is thread safe and recommended to use single Core instance. I verified with a standalone C++ test that the create_tensor for USM will return unique base pointer for each call.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
anikulk: Done

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
anikulk: Done

litert/vendors/intel_openvino/compiler/weight_bank.h
anikulk: We need to rework almost all the files. I will follow up in a separate change

taste1981 (approved):
LGTM.

litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -63,6 +63,112 @@ litert_test(
...
+    name = "weight_bank_test",
+    srcs = ["weight_bank_test.cc"],
+    copts = ["-fexceptions"],
+    data = [
+        "//litert/test:mlir_test_data",
+        "//litert/test:tflite_test_data",
+    ],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    tags = [
+        "manual",
```
MatthiasKreileder, line 88: Is it possible to actually enable some of tests in CI https://github.com/google-ai-edge/LiteRT/blob/b2b614f1b0c800032f6b297b72e679e5fc7657f2/.github/workflows/windows_x86_64.yml#L132 ?

We should build them, and where possible run them too. As the codebase grows more complex this would help a lot.

MatthiasKreileder (commented):
Please try to enable some of the tests in CI where possible and rebase to heads once more.

Thanks for the contribution.

MatthiasKreileder (changes requested):
Thanks, just some testing comments.

litert/vendors/intel_openvino/compiler/BUILD
anikulk: Done

MatthiasKreileder (approved):

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -504,6 +514,45 @@ LiteRtStatus LiteRtCompilerPluginCompile(
...
+    // serialized blob is returned for every partition; the dispatcher selects
+    // its subgraph and resolves the subgraph's weights against the shared pool.
+    //
+    // Enabled automatically when EVERY partition targets GPU (the only backend
+    // that supports the weights-as-Parameters + shared-USM dispatch path). A
+    // model with any non-GPU partition falls back to standalone baked-weights
+    // bytecode. Determining this up front (rather than per-partition inside the
+    // loop) keeps the container all-or-nothing, so we never emit a half-shared
+    // model.
+    bool share_weights = num_partitions > 0;
```
YunanAZ: Consider to add a new option to disable weight sharing.

anikulk: Done

PR #8834, opened 2026-07-21
Fix Google-Tensor AOT crash on blockwise-quantized models (freed scales tensor in OutlinePartition)

### LINKED ISSUE #8831: Google-Tensor AOT: blockwise-quantized model fails to serialize — "Failed to find scales tensor in map during serialization"
### Summary

`aot_compile` of a model with **blockwise (per-group) weight quantization** fails during
serialization of the compiled result. Reproduced with both **int2** and **int4** blockwise weights
(e.g. `ai_edge_quantizer` recipes `dynamic_wi2b32_afp32` / `dynamic_wi4b32_afp32`) targeting the
Google-Tensor backend. The compiler loads, partitions, and compiles the op fine — it fails only when
packing the compiled model back to a flatbuffer.

### Error

```
ERROR: [litert_to_flatbuffer.cc:141] Failed to find scales tensor in map during serialization
ERROR: [model_serialize.cc:558] Failed to pack as tflite
```

### Root cause

- `MapQuantizationBlockWise()` in `litert/core/model/litert_to_flatbuffer.cc` looks up the blockwise
  `.scales` (and `.zero_points`) **tensor pointer** in `tensor_map` and returns an error if it isn't
  present.
- `PackSubgraph()` in `litert/core/model/model_serialize.cc` populates `tensor_map` from **only**
  `litert_subgraph.Tensors()`.
- A blockwise weight's scales/zero_points live in a **separate data tensor** that is referenced only
  by the weight's `Qparams()` — it is **not** an op input/output. A compile-time graph copy that only
  carries op-referenced tensors drops it from the subgraph tensor list, so its pointer is never added
  to `tensor_map` → the lookup in `MapQuantizationBlockWise()` fails.

Note that `litert/core/model/shape_inference.cc` already contains logic to **re-map** blockwise
`.scales` when copying a subgraph — the Google-Tensor compile flow apparently doesn't run that path.
The failure is identical for int2 and int4, i.e. it's a serializer/graph-copy gap, not a
quantization-specific issue.

### Repro (sketch)

1. Build a small `FULLY_CONNECTED` model, quantize weights blockwise, e.g.
   `Quantizer(...).load_quantization_recipe(recipe.dynamic_wi2b32_afp32())` (or `dynamic_wi4b32_afp32`)
   → produces a tensor with `QuantizationDetails_BlockwiseQuantization`.
2. `aot_compile(...)` it for a Google-Tensor target.
3. Compile reaches `apply_plugin` and aborts with the error above.

### Proposed fix

Serializer-side robustness: in `PackSubgraph()`, before building `tensor_map`, include any
blockwise `scales`/`zero_points` tensors that are referenced by a tensor's `Qparams` but missing from
the subgraph tensor list, so they get serialized and indexed. (tflite permits unreferenced data
tensors.)

```diff
--- a/litert/core/model/model_serialize.cc
+++ b/litert/core/model/model_serialize.cc
@@
 #include "absl/container/flat_hash_map.h"  // from @com_google_absl
+#include "absl/container/flat_hash_set.h"  // from @com_google_absl
@@ LiteRtStatus PackSubgraph(SerializationContext& builder,
     tfl_subgraph.name = std::string(litert_subgraph.Name());
   }

-  for (auto* tensor : litert_subgraph.Tensors()) {
+  // Blockwise-quantized weights reference a separate scales (and optional
+  // zero_points) data tensor via their Qparams. Those tensors are not op
+  // inputs/outputs, so a compile-time graph copy can drop them from the
+  // subgraph tensor list. If we only serialize litert_subgraph.Tensors(),
+  // MapQuantizationBlockWise() cannot find the scales tensor in tensor_map.
+  // Augment the pack list with any referenced-but-missing scales/zero_point
+  // tensors.
+  std::vector<LiteRtTensor> tensors_to_pack(litert_subgraph.Tensors().begin(),
+                                            litert_subgraph.Tensors().end());
+  {
+    absl::flat_hash_set<LiteRtTensor> present(tensors_to_pack.begin(),
+                                              tensors_to_pack.end());
+    const size_t n = tensors_to_pack.size();
+    for (size_t i = 0; i < n; ++i) {
+      auto* tensor = tensors_to_pack[i];
+      if (tensor->Qparams().first != kLiteRtQuantizationBlockWise) continue;
+      const auto& bw = tensor->Qparams().second.block_wise;
+      for (LiteRtTensor extra : {bw.scales, bw.zero_points}) {
+        if (extra != nullptr && present.insert(extra).second) {
+          tensors_to_pack.push_back(extra);
+        }
+      }
+    }
+  }
+
+  for (auto* tensor : tensors_to_pack) {
     tfl_subgraph.tensors.push_back(std::make_unique<TflTensor>());
     tensor_map.insert({tensor, tfl_subgraph.tensors.size() - 1});
   }

-  for (size_t i = 0; i < litert_subgraph.Tensors().size(); ++i) {
-    auto* tensor = litert_subgraph.Tensors()[i];
+  for (size_t i = 0; i < tensors_to_pack.size(); ++i) {
+    auto* tensor = tensors_to_pack[i];
     LITERT_RETURN_IF_ERROR(
         PackTensor(builder, *tensor, *tfl_subgraph.tensors[i], tensor_map));
   }
```

**Open question for maintainers:** this assumes the blockwise `scales`/`zero_points` `LiteRtTensorT`
is still alive at serialization time (merely absent from the subgraph list). If the graph copy
actually *frees* it, the robust fix belongs upstream — have the compile-time graph copy **keep and
re-map** the blockwise scales tensor (mirroring the existing `shape_inference.cc` logic) rather than
dropping it.

### Validation

- Patch applies cleanly to current `main`.
- `@litert//litert/core/model:model_serialize` **compiles cleanly** with the patch
  (`--config=android_arm64`, bazel 7.6.1, NDK r28b): `Build completed successfully`,
  producing `libmodel_serialize.{a,so}`. Types (`Qparams()`, `block_wise.scales`/`zero_points`,
  `LiteRtTensor`, `flat_hash_set`) check out.

Happy to open a PR if this approach looks right.

Thr45hx:
Fixes #8831.

### Problem
`aot_compile` of a **blockwise (group) quantized** model on the Google-Tensor NPU backend crashes — a segfault in `apply_plugin_main` right after `Serializing model`, or (on some paths) `litert_to_flatbuffer.cc: Failed to find scales tensor in map`. Reproduces with `ai_edge_quantizer` recipes `dynamic_wi4b32_afp32` / `dynamic_wi2b32_afp32` (int4/int2 block-32). Affects Tensor G3–G6.

### Root cause
`GraphSlicer::SlicePartitionFromGraph` (`OutlinePartition`, `litert/compiler/plugin/algo.cc`) slices the selected ops into a new subgraph, cloning the ops and their **op-I/O** tensors, then calls `DCE(root)`. A blockwise-quantized weight is cloned (it's an op input), but its `scales`/`zero_points` tensors are referenced **only via `Qparams`** — not as op inputs/outputs — so they stay in `root` and are **freed by `DCE(root)`**. The cloned weight's `.scales` then dangles, and serializing the outlined partition dereferences freed memory.

### Fix
`GraphSlicer::CloneBlockwiseScales()`: before `DCE(root)`, clone any blockwise `scales`/`zero_points` tensors into the slice and remap the weight's `Qparams` pointers to the clones — mirroring the existing blockwise remap in `litert/core/model/shape_inference.cc`. Once the scales survive the copy, serialization finds them and packs the model correctly. Pure addition (no behavior change for non-blockwise models).

### Validation
Built a patched `apply_plugin_main` and ran `aot_compile` of a blockwise-quantized `FULLY_CONNECTED` model on the Google-Tensor backend:
- **Before:** segfault during serialization of the outlined partition.
- **After (`dynamic_wi4b32_afp32`, int4 block-32):** no crash — the backend compiler runs and emits a valid DGC (`AOT models: 1 failed: []`, ~35 KB output `.tflite`).

Note: `dynamic_wi2b32_afp32` (int2 block-32) gets past serialization with this fix too, but then fails *inside the backend compiler* with `INTERNAL` — a separate backend-codegen limitation (int4-blockwise is supported, int2-blockwise apparently isn't), not this graph-copy bug.

### DIFF
--- a/litert/compiler/plugin/algo.cc
+++ b/litert/compiler/plugin/algo.cc
@@ -212,6 +212,10 @@ class GraphSlicer {
 
   void RerouteTensorsThroughCustomOp(const LiteRtSubgraphT& root);
 
+  // Clones blockwise-quantization scales/zero_points tensors into the slice and
+  // remaps the Qparams pointers, so they survive DCE of the root subgraph.
+  void CloneBlockwiseScales();
+
   LiteRtSubgraph slice_;
   // Maps tensor in old subgraph to tensor in new subgraph.
   InsertOrderMap<LiteRtTensor, LiteRtTensor> tensor_map_;
@@ -257,6 +261,11 @@ LiteRtOp GraphSlicer::SlicePartitionFromGraph(
   ABSL_DCHECK(slicer.dispatch_op_->Outputs().empty());
   MakeDispatchOp(*slicer.dispatch_op_);
   slicer.RerouteTensorsThroughCustomOp(root);
+
+  // Blockwise scales/zero_points are referenced only via Qparams (not op I/O),
+  // so the op clone above leaves them in `root`; clone+remap them into the slice
+  // before DCE frees them.
+  slicer.CloneBlockwiseScales();
 
   DCE(root);
 
@@ -316,6 +325,34 @@ void GraphSlicer::CloneInto(const LiteRtOpT& old_op) {
 
     // Update the values defined in scope of the new subgraph.
     tensor_map_.InsertOrAssign(old_output, new_output);
+  }
+}
+
+void GraphSlicer::CloneBlockwiseScales() {
+  // Blockwise-quantized weights reference their `scales` (and optional
+  // `zero_points`) tensors via Qparams rather than as op inputs/outputs, so the
+  // op clone does not bring them into the slice. Clone them in and remap the
+  // Qparams pointers to the clones; otherwise they dangle when `root` is DCE'd,
+  // which crashes serialization of the outlined partition.
+  std::vector<LiteRtTensor> cloned_tensors(slice_->Tensors().cbegin(),
+                                           slice_->Tensors().cend());
+  for (auto* tensor : cloned_tensors) {
+    if (tensor->Qparams().first != kLiteRtQuantizationBlockWise) {
+      continue;
+    }
+    auto& block_wise = tensor->Qparams().second.block_wise;
+    for (LiteRtTensor* field : {&block_wise.scales, &block_wise.zero_points}) {
+      if (*field == nullptr) {
+        continue;
+      }
+      if (tensor_map_.Contains(*field)) {
+        *field = tensor_map_.Find(*field)->get().second;
+      } else {
+        auto* cloned = &MakeClone(*slice_, **field);
+        tensor_map_.InsertOrAssign(*field, cloned);
+        *field = cloned;
+      }
+    }
   }
 }
 


Thr45hx:
## Validated end-to-end on a real model: MiniCPM5-1B, full multi-subgraph int4-blockwise → compiles clean

Following up on this PR with a real LLM (not just the toy FC in the description).

**Model:** MiniCPM5-1B (`LlamaForCausalLM`, 24 layers), exported via `litert_torch` `export_hf` → 2 signatures `{prefill_128: subgraph 0, decode: subgraph 1}`.
**Quant:** `ai_edge_quantizer` weight-only **INT4 `BLOCKWISE_32`** on all FC/BMM (embed int8). float prefill+decode 3523 MB → 498 MB (7.08×).
**Target:** Google Tensor G4, `aot_compile` → `apply_plugin`.

**Before the fix** (stock `apply_plugin_main`): segfault right after `Serializing model` (the blockwise `scales`/`zero_points`, referenced only via `Qparams`, are freed by `DCE(root)` in `SlicePartitionFromGraph`).

**With `CloneBlockwiseScales`** (patched `apply_plugin_main` built from `algo.cc` + this fix):
```
Compiling Google_Tensor_G4: apply_plugin: 100%|██████████| 1/1 [07:25<00:00, 445.91s/it]
[aot] 446.0s models=1 failed=[]
[aot] DGC 532MB
RESULT=AOT_OK gran=blockwise
```
308 rank-3 FC ops across the two subgraphs, a real **532 MB DGC**, packaged into a **738 MB `.litertlm`** — no serialization crash, `failed=[]`.

So the fix holds on a full multi-subgraph blockwise model, not only the minimal repro. (This is compile/serialization validation; `int2`-blockwise still returns `INTERNAL` from the closed backend codegen, which is separate from this graph-copy bug as noted above.)

PR #8912, opened 2026-07-27
Fix handling of BATCH_MATMUL(adj_y=true)

paulbauriegel:
While testing [Whisper Large V3 Turbo](https://huggingface.co/litert-community/whisper-large-v3-turbo) with the LiteRT GPU delegate, I found multiple sources of CPU/GPU output divergence. This PR fixes one of them: the GPU parser ignored the `adj_x` and `adj_y` options of `BATCH_MATMUL`, allowing the operation to run with the wrong matrix orientation.

## Changes

- Handle both `adj_x` and `adj_y` by adding internal GPU transpose operations.
- Support all four option combinations and rank-2, rank-3, and rank-4 tensors.
- Preserve tensor type, shape, and quantization information.
- Apply the same handling to constant right-hand inputs.
- Reject unsupported tensor ranks instead of silently producing incorrect results.

### DIFF
--- a/tflite/delegates/gpu/cl/kernels/BUILD
+++ b/tflite/delegates/gpu/cl/kernels/BUILD
@@ -46,6 +46,25 @@ cc_test(
 )
 
 cc_test(
+    name = "batch_matmul_test",
+    srcs = ["batch_matmul_test.cc"],
+    linkstatic = True,
+    tags = tf_gpu_tests_tags() + [
+        "linux",
+        "local",
+    ],
+    deps = [
+        "//tflite:framework",
+        "//tflite/delegates/gpu:delegate",
+        "//tflite/kernels:test_main",
+        "//tflite/kernels:test_util",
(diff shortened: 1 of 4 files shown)


snnn (changes requested):
Thanks for the contribution. LiteRT's GPU delegate isn't open source yet.  Therefore it is hard for you to have a full picture on the problem. After reviewing the internal code, I think there is a better approach to fix the issue.

paulbauriegel:
@snnn That makes sense, feel free to close or change the MR based on what internal changes might be needed. If there is anything I can help with let me know. Otherwise just ping me if there is a new version to test 🙂

aravindmurali22:
Hi @snnn, following up on your comment above, has the internal fix for `BATCH_MATMUL` (`adj_x `/ `adj_y`) been landed or scheduled to sync to main? Thanks!

PR #9075, opened 2026-08-03
Skeleton of Arm(R) Compiler Plugin & Dispatch API

gunes-arm:
Signed-off-by: Gunes Bayir <gunes.bayir@arm.com>


litert/c/BUILD
```diff
@@ -11,6 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+#
+# SPDX-FileCopyrightText: Copyright 2026 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
MatthiasKreileder, line 15: Could you help me understand why this is needed? We didn't need this to add options for the other vendors. Here and in other files.

litert/vendors/arm/compiler/BUILD
```diff
@@ -0,0 +1,48 @@
...
+#
+
+load("//litert/build_common:litert_build_defs.bzl", "litert_dynamic_lib")
+
+package(
+    default_visibility = ["//litert:__subpackages__"],
+)
+
+litert_dynamic_lib(
+    name = "compiler_plugin",
```
MatthiasKreileder, line 13: Please enable the build of the compiler plugin and dispatch delegate on the platforms / build variants you intend to support in https://github.com/google-ai-edge/LiteRT/tree/main/.github/workflows so we have some CI that runs on PR validation.

MatthiasKreileder (commented):
Thanks for the work, just some minor questions / comments.

litert/c/BUILD
gunes-arm: Hi @MatthiasKreileder thank you very much for your comment!

Is the issue having the repetition of Apache-2.0, i.e. `SPDX-License-Identifier: Apache-2.0` or in general having the Arm copyright notice in this file, or the location of it?

If it's about having the copyright notice, as a general principle, we add our copyright notice to the files we contribute.

Is this something that you do not prefer to have? If so, can you clarify the reasons so we can discuss this and get back to you?

Many thanks.

MatthiasKreileder: Other vendor only added copyright in their `vendor/abc` folder, but not in the general/shared LiteRT files. That's why I am wondering if there's a requirement for adding it outside `vendor/arm` for you?

I am asking internally too, perhaps I am overthinking it. Hope to have a reply soon so we don't delay on this.

gunes-arm: Thanks for looking into it on your side, too. Basically, the Arm copyright was added to reflect Arm's contributions to the file.

MatthiasKreileder: Ok, so the feedback from the wider team is to please limit copyright notice to files under your vendor folder (`vendor/arm`) like the other vendors do. Would that be acceptable for you?

MatthiasKreileder: Marking as resolved, adding the copyright is OK. thanks!

litert/vendors/arm/compiler/BUILD
gunes-arm: Thanks Matthias. I've just added `build_test`inside `litert/vendors/arm/BUILD` to have an initial build at least. Does your CI pick it up?

MatthiasKreileder (approved):
Thanks!

PR #9142, opened 2026-08-05
[MLDrift] Fix FULLY_CONNECTED shared-bias handling with runtime weights

huningxin:
Register the shared bias independently from shared weights so runtime-weight fully connected nodes still get the required LINEAR bias layout. This avoids the GPU delegate compile failure seen on EfficientNet-Lite4 in the WebNN path.

Fix Chromium issue: https://issues.chromium.org/issues/534394048

### DIFF
--- a/ml_drift_delegate/tflite/model_builder.cc
+++ b/ml_drift_delegate/tflite/model_builder.cc
@@ -3222,13 +3222,17 @@ class FullyConnectedOperationParser : public TFLiteOperationParser {
       ConfigSharedBiasFullyConnectedNode(bias_share.IsShared(),
                                          tflite_node->inputs, kInputBiasId,
                                          reader, graph, node);
-      if (weights_share.IsShared()) {
+      // Register weights and bias independently so a shared bias keeps
+      // Layout::LINEAR even when weights are runtime tensors.
+      {
         auto node_inputs = graph->FindInputs(node->id);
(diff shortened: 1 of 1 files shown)


huningxin:
@snnn , PTAL, thanks!

/cc @reillyeon @lynne0326

huningxin:
@pyu10055, could you please take a look or help assign some reviewers? Thanks!

impjdi:
The bias in a CONV_2D and FULLY_CONNECTED is reserved for the learned weights that, I think, come from batch norm, and therefore constant.  If you are trying to add something with runtime weights, then that should be a regular ADD op, rather than trying to squeeze your weight into FULLY_CONNECTED.

huningxin:
> If you are trying to add something with runtime weights, then that should be a regular ADD op, rather than trying to squeeze your weight into FULLY_CONNECTED.

Sounds reasonable, thanks @impjdi . But I am wondering whether LiteRT should avoid model compiling failure for this case. 

Is it something that the `FullyConnectedOperationParser::IsSupported()` TODO tracking? https://github.com/google-ai-edge/LiteRT/blob/main/ml_drift_delegate/tflite/model_builder.cc#L3263C1-L3265C28

>    // TODO(b/372428865) : Get tensor sharing information in IsSupported.
>    //            To make checkings more completed, sharing information is
>    //            required.

PR #9151, opened 2026-08-06
Extend tensor rank support to 6D for Segment Anything models

fujunwei:
Expand the ml_drift_delegate to handle 5D and 6D tensors, which are used by Segment Anything models. Support 6D reshape by extending ConvertReshape and Handle constant tensor input in ReshapeOperationParser via AddConstInput instead of AddInput to avoid dangling consumers.


snnn:
Thanks for the contribution. We are enabling 5D tensor support in MLDrift.  It is a big change that we have been working on for months.  The details will be open soon. Currently there is no plan to support 6D tensors. The 6D tensors need to be handled by converters.

PR #9158, opened 2026-08-06
[Intel OV] Enable weight sharing for NPU

akodanka:
- Unify OpenVinoGlobalGraph into one zero-copy container
- Add NPU weightless weight sharing via temp-file staging

### DIFF
--- a/litert/vendors/intel_openvino/compiler/BUILD
+++ b/litert/vendors/intel_openvino/compiler/BUILD
@@ -110,6 +110,8 @@ litert_lib(
     features = ["-use_header_modules"],  # Incompatible with -fexceptions.
     visibility = ["//litert/vendors/intel_openvino:__subpackages__"],
     deps = [
+        "@com_google_absl//absl/log:absl_check",
+        "@com_google_absl//absl/types:span",
         "//litert/c:litert_common",
         "//litert/cc:litert_expected",
     ],
@@ -125,6 +127,7 @@ litert_test(
     # tests, which link @intel_openvino and need the OV runtime/hardware.
     deps = [
         ":global_graph",
+        "@com_google_absl//absl/types:span",
     ],
 )
 
--- a/litert/vendors/intel_openvino/compiler/BUILD
+++ b/litert/vendors/intel_openvino/compiler/BUILD
@@ -173,6 +173,58 @@ litert_test(
 )
 
 litert_lib(
+    name = "alias_shared_constants",
+    srcs = [
+        "alias_shared_constants.cc",
+        # Vendored OpenVINO dev_api header (not in the runtime SDK; symbols
+        # exported by libopenvino.so). Tags weight Constants with
+        # WeightlessCacheAttribute on the NPU shared path.
+        "weightless_caching_attributes.hpp",
+    ],
+    hdrs = ["alias_shared_constants.h"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    deps = [
+        ":weight_bank",
+        "//litert/c/internal:litert_logging",
+        "@com_google_absl//absl/types:span",
+        "@intel_openvino//:openvino",
+    ],
+)
+
+litert_test(
+    name = "alias_shared_constants_test",
+    srcs = [
+        "alias_shared_constants_test.cc",
+        "weightless_caching_attributes.hpp",
+    ],
+    copts = ["-fexceptions"],
+    data = [
+        "//litert/test:mlir_test_data",
+        "//litert/test:tflite_test_data",
+    ],
+    features = ["-use_header_modules"],  # Incompatible with -fexceptions.
+    tags = [
+        "manual",
+        "nobuilder",
+        "notap",
+    ],
+    deps = [
+        ":alias_shared_constants",
+        ":graph_iterator",
+        ":weight_bank",
+        "//litert/c:litert_runtime_c_api_shared_lib",
+        "//litert/c/internal:litert_compiler_context",
+        "//litert/cc/dynamic_runtime:litert_extended_model",
+        "//litert/compiler/cc:litert_model",
+        "//litert/test:common",
+        "//litert/test:load_test_model",
+        "//litert/test:test_models",
+        "@intel_openvino//:openvino",
+    ],
+)
+
+litert_lib(
     name = "openvino_compile_context",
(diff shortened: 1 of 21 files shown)


litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
```diff
@@ -241,6 +245,103 @@ class CustomOStreamBuf : public std::streambuf {
...
+//
+// A Constant is aliased ONLY when its current bytes are byte-identical to the
+// pool bytes for its BufferId. A mismatch means a content-altering frontend
+// transform rewrote this weight, so aliasing would feed the graph the wrong
+// data; such weights are left baked/per-partition. Comparing bytes is the
+// robust guard -- no need to enumerate which transforms alter content.
+//
+// Returns the number of Constants aliased (i.e. actually shareable). Constants
+// with no BufferId or below the element threshold stay baked.
+size_t AliasAndTagSharedConstants(
```
anikulk: Thanks for the changes, I think the tagging of constants should live in its own file related to weight_sharing_npu or similar.

litert/vendors/intel_openvino/dispatch/openvino_shared_core.h
```diff
@@ -75,6 +94,10 @@ class OpenVINOSharedCore {
       ABSL_GUARDED_BY(state_mutex_);
   std::once_flag available_devices_once_;
   std::vector<std::string> available_devices_;
+
+  // Guards the write-once temp-file staging of the shared weights bank.
+  std::mutex bank_mu_;
```
anikulk: As flagged previously by Johnny, we need to consider a case where a single process invokes inference via OV delegate on more than one model. The bank_path_ should be shared across partitions but distinct across models. ov shared code is a singleton so should not own bank_path_. Can you align with the GPU implementation and add the sync logic in device_context.

litert/vendors/intel_openvino/dispatch/openvino_shared_core.cc
```diff
@@ -43,3 +65,92 @@ const std::vector<std::string>& OpenVINOSharedCore::GetAvailableDevices() {
   });
   return available_devices_;
 }
+
+std::string OpenVINOSharedCore::GetBankPath() {
+  std::lock_guard<std::mutex> lock(bank_mu_);
+  return bank_path_;
+}
+
+std::string OpenVINOSharedCore::EnsureBankOnDisk(const void* data,
```
anikulk: Logic related to layout can live in weight_bank. We can rename some functions to differentiate GPU and NPU.

litert/vendors/intel_openvino/dispatch/openvino_shared_core.cc
```diff
@@ -43,3 +65,92 @@ const std::vector<std::string>& OpenVINOSharedCore::GetAvailableDevices() {
   });
   return available_devices_;
 }
+
+std::string OpenVINOSharedCore::GetBankPath() {
```
anikulk: This function is unused?

litert/vendors/intel_openvino/dispatch/openvino_shared_core.h
YunanAZ: +1, bank_path_ seems to belong to device_context, this will eliminate risk of collision between multiple models, and use of mutex.

nit: if mutex is strictly required, use absl::Mutex over std::mutex.

litert/vendors/intel_openvino/compiler/openvino_compiler_plugin.cc
akodanka: Done

litert/vendors/intel_openvino/dispatch/openvino_shared_core.h
akodanka: Done

litert/vendors/intel_openvino/dispatch/openvino_shared_core.cc
akodanka: Removed the unused function.

litert/vendors/intel_openvino/dispatch/openvino_shared_core.cc
akodanka: Done

litert/vendors/intel_openvino/compiler/BUILD
```diff
@@ -169,6 +172,58 @@ litert_test(
...
+    deps = [
+        ":weight_bank",
+        "//litert/c/internal:litert_logging",
+        "@com_google_absl//absl/types:span",
+        "@intel_openvino//:openvino",
+    ],
+)
+
+litert_test(
+    name = "alias_shared_constants_test",
```
MatthiasKreileder, line 196: Please enable the test here https://github.com/google-ai-edge/LiteRT/blob/5a3e5505d8984ee862c285fac3a9484d3a8dcffb/.github/workflows/windows_x86_64.yml#L178

is it possible to _run_ the test? At least building would be good.

litert/vendors/intel_openvino/dispatch/BUILD
```diff
@@ -64,6 +64,21 @@ litert_dynamic_lib(
     ] + select({
         "//litert:android": ["-Wl,-lc++abi"],
         "//conditions:default": [],
+    }) + select({
+        # Keep libstdc++ template instantiations (e.g. std::filesystem::path)
```
MatthiasKreileder: This further complicates the already non-trivial build / linking process.

Which change exactly triggered this? Is it just for `filesystem::path`? Just wondering if we can't simply avoid it?

MatthiasKreileder (changes requested):
Thanks for the work!

Please add some documentation in litert/vendors/intel_openvino for users. We should document the defaults for weight sharing.

litert/vendors/intel_openvino/dispatch/BUILD
akodanka: Yes, it's the filesystem::path . Else, we might have to hardcode the temp path to be used.

akodanka: I have modified the code to avoid the usage of std::filesystem on Linux.
This splits the temp-file staging by platform: mkstemp on POSIX, std::filesystem on Windows

MatthiasKreileder (approved):
Thanks!

anikulk (approved):

PR #9172, opened 2026-08-07
[MLDrift] Delegate TRANSPOSE with a constant data input

huningxin:
CheckGpuDelegateCompatibility requires TRANSPOSE to have exactly one runtime input, so a TRANSPOSE whose data input is also constant -- e.g. the transposed weights of a Gemm with transB -- is rejected and runs on CPU. That splits the delegated partition, and since the TFLite transpose kernel does not cache constant results the matrix is recomputed on every inference.

For TRANSPOSE that helper only checks input/output counts, so check the outputs directly when there are no runtime inputs and materialise the constant as a CONSTANT node in Parse().

After this fix WebNN EfficientNet-Lite4 model becomes fully delegated (was 119 ops on GPU, 188 on CPU), see details:
https://issues.chromium.org/issues/534394048

### DIFF
--- a/ml_drift_delegate/tflite/model_builder.cc
+++ b/ml_drift_delegate/tflite/model_builder.cc
@@ -5634,9 +5634,23 @@ class TransposeOperationParser : public TFLiteOperationParser {
                            const TfLiteNode* tflite_node,
                            const TfLiteRegistration* registration) final {
     ABSL_RETURN_IF_ERROR(CheckMaxSupportedOpVersion(registration, 9));
-    ABSL_RETURN_IF_ERROR(tflite::CheckGpuDelegateCompatibility(
-        context, tflite_node, registration));
-    ABSL_RETURN_IF_ERROR(PreCheckReadValue(context, tflite_node, 0));
+    // A TRANSPOSE with a constant data input has no runtime inputs, which
+    // CheckGpuDelegateCompatibility rejects. For TRANSPOSE that helper only
+    // checks input/output counts, so check the outputs here instead and
+    // materialise the constant in Parse().
+    const int runtime_inputs =
+        GetNumberOfRuntimeInputsForNode(context, tflite_node);
+    if (runtime_inputs == 0) {
+      if (tflite_node->outputs->size != 1) {
+        return absl::InternalError(
+            absl::StrCat("Expected 1 output tensor(s), but node has ",
+                         tflite_node->outputs->size, " output(s)."));
+      }
+    } else {
+      ABSL_RETURN_IF_ERROR(tflite::CheckGpuDelegateCompatibility(
+          context, tflite_node, registration));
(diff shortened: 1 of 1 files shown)


huningxin:
@snnn , PTAL, thanks!

/cc @reillyeon @lynne0326

huningxin:
@pyu10055, could you please take a look or help assign some reviewers? Thanks!

snnn:
In that case, should the node be const folded?

impjdi:
I agree to @snnn , it should be constant folded.  On MLD (and also the older TFL GPU), we do not try to be flexible and accommodate any kind of workload -- the NN architect should do his/her own due diligence and optimize it upfront.  We have many ops where we don't allow runtime computation that could have been precomputed.

huningxin:
> I agree to @snnn , it should be constant folded.

@impjdi @snnn , does LiteRT support constant-folding for TRANSPOSE node? Or LiteRT expect the users (WebNN) doing constant-folding?

reillyeon:
It sounds like LiteRT expects WebNN to do the necessary constant folding (or itself enforce that developers don't pass in graphs that require additional constant folding). I know WebNN attempts to do some constant folding for transposes. I don't know why it's not working in this scenario. I assume because the transpose isn't provided by the developer but is actually inserted as part of lowering from WebNN to LiteRT, a corner case in our transpose folding logic that isn't covered.

Note that transposes cannot always be eliminated by constant folding, such as if a constant is used as the input to two operators after being transposed in different ways.

huningxin:
> I assume because the transpose isn't provided by the developer but is actually inserted as part of lowering from WebNN to LiteRT, a corner case in our transpose folding logic that isn't covered.

@reillyeon , that's correct, the Transpose is inserted by WebNN LiteRT backend when gemm.b_transpose is false (default): https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/tflite/graph_builder_tflite.cc;l=6190

I can look into transposing it when building the LiteRt graph. 

> Note that transposes cannot always be eliminated by constant folding, such as if a constant is used as the input to two operators after being transposed in different ways.

Those cases would fall back to CPU.

snnn:
Transpose and reshape are similar in this case. I can share some thoughts about how to optimize them way: 
1. remove all transpose nodes from the graph and tag the uses, which means that each edge on the model graph carries a "layout" tag that tells if the tensor needs to be transposes before use.
2.  flow the tags in the graph and try to eliminate unnecessary tags. For example, one edge can only have one tag. Therefore two adjacent transposes/reshapes can be combined.
3.  apply the tags back to graph by adding transpose nodes

This algorithm can handle the multi-use case correctly. i.e.  if a constant is used as the input to two operators after being transposed in different ways.

The basic idea is: in a model graph (which is a DAG). Every node is an operation(such as mul, add) and every edge is a use. An edge can have a tag to express if a reshape/transpose is needed.  The tag is optional.

PR #9216, opened 2026-08-11
[Intel_OpenVINO] Log OpenVINO execution and version details for Intel NPU dispatch

jatinwadhwa921:
## Summary

Improve Intel OpenVINO dispatch diagnostics so LiteRT users can clearly see which device is selected for each compiled partition.

## Details

When a model is partitioned and compiled for OpenVINO, the dispatch layer now logs the device selected for each partition, such as CPU, GPU, or NPU. This makes it easier to verify where each part of the model is running.

These logs also help avoid confusion caused by unrelated runtime messages, such as XNNPACK delegate creation. The presence of an XNNPACK message does not necessarily mean that inference is running on the CPU. The new OpenVINO dispatch logs provide a direct indication of the device actually selected for each compiled partition.

## Motivation

This change makes device selection easier to understand when debugging or validating LiteRT inference.

## Testing

- Verified that the diagnostic logs identify the selected device for each compiled partition.
- No inference behavior or device-selection logic was changed.
- This change only improves runtime diagnostics.


litert/vendors/intel_openvino/dispatch/invocation_context.cc
```diff
@@ -51,6 +54,24 @@
 
 namespace {
 
+void LogNpuCompilerVersionOnce(const ov::Core& core) {
```
akodanka: Since this is needed only once, can we print this from DispatchInitialize ?
That can avoid the logic for logging once.

litert/vendors/intel_openvino/dispatch/invocation_context.cc
```diff
@@ -296,6 +317,43 @@ LiteRtDispatchInvocationContextT::Create(
     return litert::Error(kLiteRtStatusErrorRuntimeFailure, e.what());
   }
 
+  if (device == "NPU") {
+    LogNpuCompilerVersionOnce(*core);
+  }
+
+  try {
```
akodanka: Shoundn't the device print at Line 290 be sufficient, you may append the function_name to that log.
When using an AOT model, the device_from_header is expected to capture that already.

akodanka (changes requested):
Thanks @jatinwadhwa921 
Please find a few suggestions.

litert/vendors/intel_openvino/dispatch/invocation_context.cc
jatinwadhwa921: fixed

litert/vendors/intel_openvino/dispatch/invocation_context.cc
jatinwadhwa921: But, i thought this would bring extra clarity to the user at dispatch time.

litert/vendors/intel_openvino/dispatch/dispatch_api.cc
```diff
@@ -14,12 +14,16 @@
 // limitations under the License.
 
 #include <cstddef>
+#include <cstdint>
+#include <exception>
```
akodanka: Can we use ov::Exception instead of std::exception.

Refer https://github.com/google-ai-edge/LiteRT/commit/1ce6ea0490aee72bcfd5754327e71d94faff902d

litert/vendors/intel_openvino/dispatch/invocation_context.cc
```diff
@@ -220,17 +220,9 @@ LiteRtDispatchInvocationContextT::Create(
     exec_bytecode_ptr =
         static_cast<const uint8_t*>(exec_bytecode_ptr) + payload_offset;
     exec_bytecode_size -= payload_offset;
-    LITERT_LOG(LITERT_INFO, "Dispatch: using device '%s' from bytecode header",
```
akodanka, line 223: Wouldn't these logs still be useful to detect where the device selection came from ?

litert/vendors/intel_openvino/dispatch/dispatch_api.cc
```diff
@@ -111,11 +115,34 @@ LiteRtStatus DispatchInitialize(const LiteRtRuntimeContext* runtime_context,
...
+
   ov::Core core;
-  std::vector<std::string> availableDevices = core.get_available_devices();
-  for (auto&& device : availableDevices)
+  std::vector<std::string> available_devices = core.get_available_devices();
+  bool npu_available = false;
+  for (const auto& device : available_devices) {
     LITERT_LOG(LITERT_INFO, "[Openvino]Found device plugin for: %s",
                device.c_str());
+    npu_available = npu_available || device == "NPU" ||
```
akodanka: nit : Would a single check for be sufficient ?
`npu_available || device.rfind("NPU", 0)`

jatinwadhwa921: fixed

litert/vendors/intel_openvino/dispatch/invocation_context.cc
jatinwadhwa921: Fixed

litert/vendors/intel_openvino/dispatch/dispatch_api.cc
jatinwadhwa921: Fixed

PR #9464, opened 2026-08-20
Fix L2 normalization fusion for multi-axis reductions

### LINKED ISSUE #9212: tf.math.l2_normalize with default axis normalizes per-last-axis instead of over the whole tensor like eager TensorFlow
## Environment

- LiteRT: `ai-edge-litert 2.1.6`
- TensorFlow converter: `tf-nightly 2.22.0.dev20260808` (also reproduces bit-identically on `tf-nightly 2.22.0.dev20260726` and on TensorFlow 2.20.0)
- OS: Linux x86_64, CPU
- Python: 3.11

## Minimal reproduction

```python
import tensorflow as tf
from ai_edge_litert.interpreter import Interpreter, OpResolverType


class Model(tf.keras.Model):
    def call(self, x):
        return tf.math.l2_normalize(x)


x = tf.constant([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]], dtype=tf.float32)
model = Model()
expected = model(x).numpy()
flatbuffer = tf.lite.TFLiteConverter.from_keras_model(model).convert()


def run(**kwargs):
    interpreter = Interpreter(model_content=flatbuffer, **kwargs)
    input_detail = interpreter.get_input_details()[0]
    interpreter.resize_tensor_input(input_detail["index"], list(x.shape))
    interpreter.allocate_tensors()
    input_detail = interpreter.get_input_details()[0]
    interpreter.set_tensor(input_detail["index"], x.numpy())
    interpreter.invoke()
    return interpreter.get_tensor(interpreter.get_output_details()[0]["index"])


print("eager:", expected)
print("LiteRT default:", run())
print(
    "LiteRT without default delegates:",
    run(experimental_op_resolver_type=OpResolverType.BUILTIN_WITHOUT_DEFAULT_DELEGATES),
)
```

## Actual result

```text
eager: [[0.10482848 0.20965695 0.31448543]
 [0.4193139  0.5241424  0.62897086]]
LiteRT default: [[0.26726124 0.5345225  0.8017837 ]
 [0.45584232 0.5698029  0.68376344]]
LiteRT without default delegates: [[0.26726124 0.5345225  0.8017837 ]
 [0.45584232 0.5698029  0.68376344]]
```

`tf.math.l2_normalize(x, axis=None)` in eager TensorFlow normalizes over the
entire tensor (the L2 norm of all 6 elements, `sqrt(91) ≈ 9.539`, is used as
the denominator for every element). The converted LiteRT model instead
normalizes each row independently using only that row's own L2 norm
(`sqrt(14)` for row 0, `sqrt(77)` for row 1) — this is the behavior of
`axis=-1`, not `axis=None`. Both the default XNNPACK-delegated interpreter and
the interpreter without default delegates produce the same wrong result, so
this looks like a converter/graph-construction issue rather than a
delegate-specific kernel bug.

## Expected result

The converted model should match eager TensorFlow's whole-tensor
normalization for the default `axis=None`.

## Additional context

This is the same root cause as tensorflow/tensorflow#118181 and its duplicate
#120292 (both used a `[3, 1]` input, where the wrong per-row normalization of
a single-element "row" happens to produce exactly `[1.0, 1.0, 1.0]`). Both
were closed on 2026-06-09/2026-06-17 with a note that active TFLite
development has moved to LiteRT and to refile there if the issue persists —
they were not closed because the bug was fixed. I confirmed with the
reproducer above that this is still present as of
`tf-nightly 2.22.0.dev20260808` (bit-identical to `2.22.0.dev20260726` from
two weeks earlier), and could not find an existing LiteRT-repo report, so I'm
refiling it here per that guidance.

laolvfan:
Fixes #9212

## Summary
- Only fuse to `TFL_L2_NORMALIZATION` for a single trailing axis or `-1`.
- Keep all-axis reductions unfused because they normalize the whole tensor, while `TFL_L2_NORMALIZATION` only normalizes the final dimension.
- Add an MLIR regression test for the rank-2 `[0, 1]` reduction produced by `tf.math.l2_normalize(axis=None)`.

## Testing
- Reproduced the issue on `tf-nightly 2.22.0.dev20260809` with `ai-edge-litert 2.1.6`; default and builtin-only LiteRT both differed from eager TensorFlow.
- Attempted `bazel test --test_output=errors //tflite/converter/tests:optimize.mlir.test` locally. The test did not reach compilation because Bazel external-repository downloads from GitHub were too slow in this environment; the PR CI should run the target normally.

### DIFF
--- a/tflite/converter/transforms/optimize_pass.cc
+++ b/tflite/converter/transforms/optimize_pass.cc
@@ -117,25 +117,15 @@ ElementsAttr ReshapeNCHWBiasToNHWC(Value v, Attribute a) {
 }
 
 bool L2NormalizeReduceAxis(Value sq_op, DenseElementsAttr axis) {
-  if (axis.getNumElements() == 0) {
+  // TFL_L2_NORMALIZATION only reduces the final dimension. Do not fuse a
+  // multi-axis reduction: in particular, [0, 1, ..., rank - 1] reduces the
+  // whole tensor and is not equivalent for rank greater than one.
+  if (axis.getNumElements() != 1) {
     return false;
   }
-  if (mlir::cast<ShapedType>(sq_op.getType()).getRank() - 1 ==
-          *axis.getValues<int>().begin() ||
-      *axis.getValues<int>().begin() == -1) {
-    return true;
-  }
-  if (mlir::cast<ShapedType>(sq_op.getType()).getRank() !=
-      axis.getNumElements()) {
-    return false;
-  }
-  auto shape = mlir::cast<ShapedType>(sq_op.getType());
-  SmallVector<int, 4> elems{axis.getValues<int>().begin(),
-                            axis.getValues<int>().end()};
-  for (int i = 0; i < shape.getRank(); ++i) {
-    if (i != elems[i]) return false;
-  }
-  return true;
+  const int reduction_axis = *axis.getValues<int>().begin();
+  const int rank = mlir::cast<ShapedType>(sq_op.getType()).getRank();
+  return reduction_axis == rank - 1 || reduction_axis == -1;
 }
 
 // Checks if a ReshapeOp is equivalent to a `keep_dims=true` reduction by


laolvfan:
@googlebot rescan

laolvfan:
> *This was generated by my claude during triage.*


Closing this PR because I plan to submit the fix through the TensorFlow repository instead. Thank you.

PR #9521, opened 2026-08-24
Fix run_model GPU latency measurement

### LINKED ISSUE #9453: [run_model][GPU] Reported inference time measures OpenCL enqueue instead of completed execution
## Summary

`litert run ... --android --gpu` can report physically implausible sub-millisecond inference times for a fully delegated OpenCL graph because `run_model` does not enable the GPU delegate's benchmark/completion-wait mode before timing `CompiledModel::Run()`.

On an SM8550 Android device, a 366-node FP16 pose model was fully delegated to `LITERT_CL`, but `run_model` reported 274 microseconds. Enabling GPU benchmark mode in `run_model` caused the same command to report 16–22 ms, matching the dedicated `benchmark_model` result.

## Environment

- LiteRT source commit `9e377c6` (2026-08-19)
- `litert-cli-nightly==0.2.0.dev20260817`
- `ai-edge-litert-nightly==2.3.0.dev20260819`
- Android 16
- Samsung `SM-F9460`
- Qualcomm `SM8550`
- LiteRT OpenCL GPU accelerator
- FP16 TFLite model, approximately 6.46 MiB, input `1x360x640x3`

The model is proprietary and cannot be attached, but the timing defect is independent of output values and occurs after full GPU delegation.

## Reproduction

```bash
litert run CoreMLPoseDetector.tflite --android --gpu
```

Delegation succeeds:

```text
Replacing 366 out of 366 node(s) with delegate (LITERT_CL) node,
yielding 1 partitions for subgraph 0 (main).
```

Unmodified timing output:

```text
Run model for 1 times
First run took 274 microseconds
Slowest run took 274 microseconds
Fastest run took 274 microseconds
All runs took average 274 microseconds
```

The 274 microseconds are consistent with submitting work to the OpenCL queue, not completing this graph.

## Source-level difference

The dedicated benchmark path explicitly enables completion waiting:

```cpp
// Enable benchmark mode to run clFinish() after each inference.
gpu_options.EnableBenchmarkMode(/*enabled=*/true);
```

`litert/tools/run_model.cc` configures GPU precision and kernel batch size but does not enable that mode before measuring each `compiled_model.Run(...)` call with `NowMicros()`.

## Confirmed control

Adding the following to the GPU options in `run_model`:

```cpp
gpu_opts.EnableBenchmarkMode(/*enabled=*/true);
```

produces this delegate log:

```text
Benchmark mode is enabled.
Replacing 366 out of 366 node(s) with delegate (LITERT_CL) node,
yielding 1 partitions for subgraph 0 (main).
```

The exact CLI path then reports completed-execution timings:

```text
Run model for 3 times
First run took 22253 microseconds
Slowest run took 22253 microseconds
Fastest run took 16592 microseconds
All runs took average 18538 microseconds
```

The dedicated LiteRT benchmark on the same model and device reports:

```text
Inference (avg): 17.21 ms (5 runs)
Inference (min): 16.84 ms
Inference (max): 17.70 ms
```

The agreement between the synchronized `run_model` result and `benchmark_model` confirms that the original value was not completed inference latency.

## Expected behavior

Any latency printed by `run_model` should cover completed inference work. If the tool intentionally measures asynchronous submission instead, it should label that value as submission/enqueue latency rather than model execution time.

## Potential changes

1. **Enable GPU benchmark mode when `run_model` prints per-invocation timings.** Call `GpuOptions::EnableBenchmarkMode(true)` whenever the GPU accelerator is selected.
2. **Make synchronization semantics explicit.** If asynchronous submission remains useful, add a flag such as `--gpu_wait_for_completion`, default it to true for timing output, and label unsynchronized results as enqueue latency.
3. **Share GPU timing configuration.** Reuse one helper between `run_model` and `benchmark_litert_model` so the two tools cannot silently diverge on completion waiting.
4. **Clarify the synchronous API contract.** Document whether `CompiledModel::Run()` guarantees device completion for each accelerator or only host-side submission. Timing tools should add an explicit wait if the API does not guarantee completion.
5. **Print the active timing mode.** Report whether GPU completion synchronization is enabled alongside the timing summary.
6. **Add regression coverage.** At minimum, test that the GPU options created by `run_model` enable benchmark mode. An Android OpenCL integration test could also compare `run_model` latency against `benchmark_model` within a broad tolerance for a nontrivial graph.
7. **Separate warmup from measured iterations.** This is not required for correctness of the wait, but it would prevent first-run compilation/cache effects from being presented as steady-state latency.

## Investigation attribution

The reproduction, source diagnosis, patched Android build, and on-device before/after verification were performed by **OpenAI Codex (`openai-codex/gpt-5.6-sol`)** in the reporter's development environment. The reporter supplied the original implausible timing and authorized publication.

Jacfger:
## Summary

- enable GPU benchmark mode when `run_model` selects the GPU
- propagate failure to configure benchmark mode
- reuse the selected `GpuOptions` instance for the existing kernel-batch and precision configuration

## Problem

[`CompiledModel::Run()`](https://github.com/google-ai-edge/LiteRT/blob/9e377c6/litert/cc/litert_compiled_model.h#L721-L728) is documented as running synchronously and, through RunHelper() and RunCApiHelper(), dispatches to the synchronous [`LiteRtRunCompiledModel`](https://github.com/google-ai-edge/LiteRT/blob/9e377c6/litert/c/litert_compiled_model.h#L126-L141) C API has the same synchronous contract.

`run_model` times each `Run()` call with host-side timestamps and reports the result as inference latency. However, GPU benchmark mode defaults to false. In the observed OpenCL path, the call returned after command submission rather than completed GPU execution, so a fully delegated 366-node pose model was reported as taking 274 microseconds even though completed inference took approximately 17 milliseconds.

The dedicated benchmark already [enables benchmark mode to run `clFinish()` after each inference](https://github.com/google-ai-edge/LiteRT/blob/9e377c6/litert/tools/benchmark_litert_model.cc#L165-L167).

## Fix

Enable `GpuOptions::EnableBenchmarkMode(true)` before constructing the compiled model. This makes the OpenCL backend finish queued work before the timed synchronous `Run()` call returns, so `run_model` reports completed inference latency rather than enqueue latency.

## Verification

Android ARM64 build:

```text
Target //litert/tools:run_model up-to-date
Build completed successfully
```

Samsung SM-A7050, Android 11, Qualcomm GPU, 366/366 nodes delegated to `LITERT_CL`:

```text
run_model average:       328.297 ms (5 runs)
benchmark_model average: 327.580 ms (5 runs)
difference:              approximately 0.22%
```

Samsung SM-F9460, Android 16, Qualcomm GPU, 366/366 nodes delegated to `LITERT_CL`:

```text
run_model average:       18.538 ms (3 runs)
benchmark_model average: 17.210 ms (5 runs)
```

Both patched runs logged `Benchmark mode is enabled.` and completed successfully.

Fixes #9453

## Submission attribution

This issue was diagnosed and this PR was implemented, built, tested on the reporter's devices, and submitted by **OpenAI Codex (`openai-codex/gpt-5.6-sol`)** in @Jacfger's development environment. @Jacfger supplied the original reproduction, model, and devices and authorized submission.

### DIFF
--- a/litert/tools/run_model.cc
+++ b/litert/tools/run_model.cc
@@ -233,22 +233,22 @@ Expected<Options> GetOptions() {
   options.SetHardwareAccelerators(accelerators);
 
   if (accelerators & litert::HwAccelerators::kGpu) {
+    LITERT_ASSIGN_OR_RETURN(auto& gpu_opts, options.GetGpuOptions());
+    // Enable benchmark mode to run clFinish() after each inference.
+    LITERT_RETURN_IF_ERROR(gpu_opts.EnableBenchmarkMode(/*enabled=*/true));
+
     int kernel_batch_size = absl::GetFlag(FLAGS_kernel_batch_size);
     if (kernel_batch_size > 0) {
-      LITERT_ASSIGN_OR_RETURN(auto& gpu_opts, options.GetGpuOptions());
       LITERT_RETURN_IF_ERROR(gpu_opts.SetKernelBatchSize(kernel_batch_size));
     }
     std::string gpu_precision = absl::GetFlag(FLAGS_gpu_precision);
     if (gpu_precision == "fp32") {
-      LITERT_ASSIGN_OR_RETURN(auto& gpu_opts, options.GetGpuOptions());
       LITERT_RETURN_IF_ERROR(
           gpu_opts.SetPrecision(GpuOptions::Precision::kFp32));
     } else if (gpu_precision == "fp16") {
-      LITERT_ASSIGN_OR_RETURN(auto& gpu_opts, options.GetGpuOptions());
       LITERT_RETURN_IF_ERROR(
           gpu_opts.SetPrecision(GpuOptions::Precision::kFp16));
     } else if (gpu_precision == "auto") {
-      LITERT_ASSIGN_OR_RETURN(auto& gpu_opts, options.GetGpuOptions());
       LITERT_RETURN_IF_ERROR(
           gpu_opts.SetPrecision(GpuOptions::Precision::kDefault));
     } else {


Jacfger:
@shivvaam0001 I've created the PR as you suggested, and I've just signed the CLA now as well.

I'm not sure if you need to manually approve anything on your end. Please let me know if you have any concerns about the PR — I'm happy to follow up on them.

PR #9621, opened 2026-08-27
Ov version update

akodanka:
Problem Statement:
The Openvino version is currently duplicated in openvino.bzl and setup.py

Solution:
Use a single source of truth : openvino_version.bzl
Introduce a dev script that generates, validates and updates the versions to it: update_openvino_version.py
Handle the nightly and release packages separately for generating the paths

Note:
This also updates the Openvino version to the latest release package: 2026.3.1

Tests done :
1. Verified the update_openvino_version.py script with nightly and release packages
2. Verified the openvino version against the symlinks generated by bazel fetch
3. Verified the OPENVINO_NATIVED_DIR overrides the configured openvino version
4. pip install from the Intel SDK sdist and verified expected openvino requirement


MatthiasKreileder (approved):
Thanks for the change and also documenting the process!

PR #9664, opened 2026-08-31
[MLDrift] Fix BHWC layout of low-rank constant operands in elementwise ops

wangw-1991:
## Problem
A non-float constant operand of a binary elementwise op was read with the default layout, i.e. TFLite dim 0 was mapped onto BHWC's batch. TFLite/numpy broadcasting is right-aligned, so a constant of lower rank than the runtime operand has to land on the *trailing* BHWC axes instead. Reading `[C]` as `Cx1x1x1` rather than `1x1x1xC` makes the GPU elementwise broadcast read batch-0 only and splat it across every channel, so the op silently produces a wrong result instead of failing to delegate.

## Fix
1. `ElementwiseOperationParser::ParseInputsWithConstTensor()` and `litert::ml_drift::ir::ParseInputsWithConstTensor` now derives that layout from the rank of both operands via a new `BroadcastConstLayout()` helper, instead of passing an empty layout. The helper encodes the same right-alignment rule as the existing `ExtractTensorShapeWithTfLiteBroadcast()`.
2. Add related tests.

see chromium issue: https://issues.chromium.org/issues/534893138

PR #9666, opened 2026-08-31
[MLDrift] Support GATHER indices of rank > 1 in the ml-drift delegate

wangw-1991:
## Problem
GATHER previously required strictly 1D indices. This adds support for indices of higher rank whose leading dimensions are all 1 (e.g. `[1,N]`, common after a reshape upstream): they hold the same vector the kernel reads, so the delegate can now relabel them instead of rejecting the op and falling back to CPU.

Because such indices give the TFLite output a higher rank than the input, the gathered dimension lands on a different BHWC axis — gathering rows of a `[V,C]` table with `[1,N]` indices yields `[1,N,C]`, not `[N,C]`. The parser now emits the GATHER with the shape the kernel produces and appends a RESHAPE to reach the real output layout.

see chromium issue: https://issues.chromium.org/issues/534893138

### DIFF
--- a/ml_drift_delegate/tflite/BUILD
+++ b/ml_drift_delegate/tflite/BUILD
@@ -241,6 +241,7 @@ cc_test(
         "@com_google_googletest//:gtest_main",
         "@ml_drift//ml_drift/common:data_type",
         "@ml_drift//ml_drift/common:model",
+        "@ml_drift//ml_drift/common:operations",
         "@ml_drift//ml_drift/common:shape",
         "@ml_drift//ml_drift/common:tensor",
         "@ml_drift//ml_drift/common:types",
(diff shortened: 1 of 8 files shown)


wangw-1991:
@snnn, could you please take a look or help assign some reviewers? Thanks!

huningxin:
@wangw-1991 , it seems you pushed wrong content, the changes don't touch gather op. Could you please double check?

wangw-1991:
> @wangw-1991 , it seems you pushed wrong content, the changes don't touch gather op. Could you please double check?

sorrry, this PR is based on https://github.com/google-ai-edge/LiteRT/pull/9664. I forgot to push gather related changes. Please only check "Add gather change." commit.

huningxin:
@wangw-1991 , I uploaded a Chromium CL which reshapes GATHER indices to 1-D for ML Drift delegate: https://chromium-review.googlesource.com/c/chromium/src/+/8354248. Could you please verify if it fixes GATHER in whisper-base models?

/cc @reillyeon

wangw-1991:
> @wangw-1991 , I uploaded a Chromium CL which reshapes GATHER indices to 1-D for ML Drift delegate: https://chromium-review.googlesource.com/c/chromium/src/+/8354248. Could you please verify if it fixes GATHER in whisper-base models?
> 
> /cc @reillyeon

Thanks, the fix in Chromium works. So I will close this issue for now.

PR #9682, opened 2026-08-31
Accept a float16 GATHER input feeding a float32 output

huningxin:
GATHER required its input and output tensors to have identical types, so a float16 table feeding a float32 destination was rejected and the node ran on CPU, splitting the delegated partition.

A framework, e.g. Chromium WebNN LiteRT backend [1], emits fp16 weights as a float16 constant behind a DEQUANTIZE, this delegate folds the DEQUANTIZE away, and the GATHER is then left reading a float16 table into the float32 destination the graph declared.

[1]: https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/tflite/graph_builder_tflite.cc;l=1288;drc=f71ffcd5aa26177023677965711de7f223b41d42

### DIFF
--- a/ml_drift_delegate/tflite/model_builder.cc
+++ b/ml_drift_delegate/tflite/model_builder.cc
@@ -3495,7 +3495,17 @@ class GatherOperationParser : public TFLiteOperationParser {
     ABSL_RETURN_IF_ERROR(
         PreGetOutputTensor(context, tflite_node, 0, &tfl_output));
     if (tfl_input->type != tfl_output->type) {
-      return absl::InvalidArgumentError("Input / output dtype mismatch.");
+      // A folded fp16 weight (e.g. Chromium emits fp16 constants behind a
+      // DEQUANTIZE that this delegate folds away) legitimately feeds an fp32
+      // destination. The kernel converts to the destination type, so allow a
+      // float/float difference; anything else is still a real mismatch.
+      const bool both_float = (tfl_input->type == kTfLiteFloat16 ||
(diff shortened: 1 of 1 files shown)


roserg:
It should not be resolved in kernel.
Gather kernel works with identical src/dst types and this should not be changed.
Instead, explicit cast operation must be added.
For example, fp16 in -> gather -> fp16 out -> cast -> fp32 out
you also can cast input, but output is preferred, it will be then automatically fused in single kernel.

reillyeon:
In the Chromium TFLite graph generator GATHER needs to be marked as supporting fp16 so that it doesn't generate the DEQUANTIZE node and instead puts it after the GATHER as suggested.

@roserg, is there a list of operators that explicitly support fp16 input like this? Chromium's TFLite graph generator has been written under the assumption that almost no TFLite ops do but that accelerators will execute the graph as if all its fp32 tensors were fp16 if the fp16 precision option is provided at a graph level.

impjdi:
`GATHER` is essentially just a lookup operation.  If looked up value is `bool`, the output should be `bool` and not `int64`, because someone downstream needs it that way.  That's the job of the `CAST` operation.

If you need to add different dtypes to be marked as compatible for a node, maybe you want to add that in the successor `IrModel` (look twhere we have dedicated checks for those litert/ml_drift/tflite/support/support_gather.cc

huningxin:
Thanks all for the feedback! I've submitted a CL to make GATHER as float16-capable on Chromium side: https://chromium-review.googlesource.com/c/chromium/src/+/8345429

PR #9712, opened 2026-09-02
[OpenVINO] Skip transformed weights in GPU sharing

zlma7001:
Only promote constants whose size and contents match the original weight bank buffer. This prevents repacked i2-to-i4 GPU weights from being bound to incompatible shared storage.

Test: compile and infer gemma4 E2B with OV-GPU manually


MatthiasKreileder (approved):
Thanks for the fix!

PR #9721, opened 2026-09-02
SLICE: validate static-output slice bounds without changing allocation type

endorphin13:
Redirected here from **[tensorflow/tensorflow#126473](https://github.com/tensorflow/tensorflow/pull/126473)**. @dmiltr3 verified it there and then asked for it to be re-submitted upstream:

> TFLite C/C++ operator kernel implementations and tests (under `tensorflow/lite/kernels/**`, including `tensorflow/lite/kernels/slice.cc` and `tensorflow/lite/kernels/slice_test.cc`) are maintained upstream in the google-ai-edge/LiteRT repository and vendored into Google3 from LiteRT. Because changes to these kernel files cannot be imported into Google3 directly from tensorflow/tensorflow, please submit this pull request against the upstream LiteRT repository.

Same change, rebased onto LiteRT `main` (files under `tflite/kernels/` here). I confirmed `tflite/kernels/slice.cc` on `main` is byte-identical to the TensorFlow copy apart from include paths, and the defect below is present verbatim at lines 146-150.

## The defect

`Prepare()` returns early when the output shape is fully specified:

```c
// If the shape of output is fully specified then resize even if
// the input shape is not staticly defined.
if (!HasUnspecifiedDimension(output) && ShapeHasRank(output->dims)) {
  return kTfLiteOk;
}
// Postpone allocation of output if any of the indexing tensors is not
// constant, or the input tensor has dynamic dimension.
if (!(IsConstantOrPersistentTensor(begin) && IsConstantOrPersistentTensor(size)) ||
    HasUnspecifiedDimension(input)) {
  SetTensorToDynamic(output);
  return kTfLiteOk;
}
```

That path returns without calling `SetTensorToDynamic(output)` and without validating anything. Note that the constancy check for `begin`/`size` sits **below** the return, so it is never reached for a statically shaped output. `Eval()` re-runs `ResizeOutputShape()` only when the output is dynamic, so `CalculateOutputShapeVector()` — the single place `begin` and `size` are checked against the input — never runs.

Two cases reach the kernel unvalidated:

1. the input extent is only known at run time; and
2. `begin`/`size` are runtime tensors, whose values `Prepare()` cannot see at all.

## Why this differs from the earlier attempt

An earlier revision of this fix ([tensorflow/tensorflow#126177](https://github.com/tensorflow/tensorflow/pull/126177)) widened the early-return condition so that models with a dynamic input dimension fell through and had their output reclassified as dynamic. That changed the allocation type of otherwise valid models, which affects arena planning and delegate behaviour. It also only addressed case 1 — case 2 above still took the early return.

This revision separates validation from resizing:

- `ValidateSliceAgainstOutput()` computes the shape implied by the current input and indices and confirms it matches the already allocated output. It never resizes the output or changes its allocation type.
- `Prepare()` keeps its early return for a statically shaped output. When the input extent and both index tensors are already known it validates in place; otherwise the check is deferred.
- `Eval()` validates the static-output path instead of skipping it, and still resizes only for a genuinely dynamic output.

**The number of `SetTensorToDynamic()` calls is unchanged from `main` (exactly one, in the same place), so no model changes allocation type as a result of this patch.**

## Hardening `CalculateOutputShapeVector()`

It becomes the validation primitive here, so two existing weaknesses are addressed:

- It indexes `begin` and `size` by input dimension, but `Prepare()` only checks `NumElements(begin) == NumElements(size)` — never against the input rank. An index vector shorter than the rank is read past its end. A rank check is added before either buffer is dereferenced.
- The bounds test was `SizeOfDimension(input, idx) < begin[idx] + size_value`. `begin` was not range-checked first, and the addition can overflow. `begin` is now validated before use, and the comparison is written as `size_value > dim_value - begin_value`.

## Tests

- a static output keeps its static allocation (guards against the #126177 regression);
- runtime `begin`/`size` out of bounds with a static output — int32 and int64;
- negative `begin`;
- index vector shorter than the input rank;
- computed shape disagreeing with the declared static output.

Not compile-tested locally (no Bazel environment) — please run CI. On the TensorFlow PR the reviewer reported `//tensorflow/lite/kernels:slice_test` compiling and passing cleanly, with all non-Windows CI platforms green.

### DIFF
--- a/tflite/kernels/slice.cc
+++ b/tflite/kernels/slice.cc
@@ -58,20 +58,33 @@ TfLiteStatus CalculateOutputShapeVector(TfLiteContext* context,
                                         const TfLiteTensor* begin,
                                         const TfLiteTensor* size,
                                         std::vector<int>* output_shape_vector) {
-  for (int idx = 0; idx < NumDimensions(input); ++idx) {
+  const int input_dims = NumDimensions(input);
+  // `begin` and `size` are indexed by input dimension below. Prepare() only
+  // checks that they agree with each other, which does not bound them against
+  // the input rank, so establish that here before either is dereferenced.
+  TF_LITE_ENSURE_EQ(context, NumElements(begin), input_dims);
+  TF_LITE_ENSURE_EQ(context, NumElements(size), input_dims);
+  for (int idx = 0; idx < input_dims; ++idx) {
+    const T dim_value = static_cast<T>(SizeOfDimension(input, idx));
+    const T begin_value = GetTensorData<T>(begin)[idx];
     T size_value = GetTensorData<T>(size)[idx];
+    // Reject an out-of-range `begin` before it is used in any arithmetic, so
+    // the bounds comparison below cannot be reached with a wrapped value.
+    if (begin_value < 0 || begin_value > dim_value) {
+      TF_LITE_KERNEL_LOG(context, "Invalid begin.");
+      return kTfLiteError;
+    }
     if (size_value < 0) {
       if (size_value != -1) {
         TF_LITE_KERNEL_LOG(context, "Invalid size.");
         return kTfLiteError;
       }
-      size_value = SizeOfDimension(input, idx) - GetTensorData<T>(begin)[idx];
-    } else {
-      if (SizeOfDimension(input, idx) <
-          GetTensorData<T>(begin)[idx] + size_value) {
-        TF_LITE_KERNEL_LOG(context, "Invalid begin and size.");
-        return kTfLiteError;
-      }
+      size_value = dim_value - begin_value;
+    } else if (size_value > dim_value - begin_value) {
+      // Written as a subtraction rather than `begin_value + size_value >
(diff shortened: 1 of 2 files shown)


snnn (changes requested):
Thanks for the contribution. 

We don't have a use case that TFLite needs to read a malicious model from untrusted source with malicious tensor shapes there.  Such validations should be put at a higher level that is backend independent, otherwise we need to duplicate this logic again and again in every backend which isn't desirable. 

Please be aware that TFLite's security model is documented at https://github.com/google-ai-edge/LiteRT/blob/main/g3doc/instructions/security.md .
If you found a case that Chromium can generate such a malicious TFLite model  in its webnn engine(https://chromium.googlesource.com/chromium/src/+/lkgr/services/webnn/) with incorrect  data types or tensor shapes, please submit a bug report to Chromium. They would redirect the issue to us if they think the issue should be fixed in LiteRT. 

You may also require us changing the security model if there is a valid use case(with real user impact) that requires that.

PR #9852, opened 2026-09-09
[MLDrift] Fix GATHER with 1-D constant indices

huningxin:
The gather kernel reads its indices along the channels axis, but a 1-D indices tensor [N] is auto-expanded to {N,1,1,1}, putting N on the batch axis, where the kernel cannot see it. ConvertGather() already reshapes 1-D *runtime* indices to {1,1,1,N} for exactly this reason; constant indices took the AddConstInput() branch with an empty layout and got no equivalent, so the kernel addressed a channels extent of 1 and failed to build.

Request Layout::SCALAR for 1-D constant indices -- the constant equivalent of the RESHAPE the runtime branch inserts. Depends on the preceding change, which makes the requested layout reach the uploaded data as well as the IrTensor.

 Also apply the requested constant layout to the tensor data. AddConstInput() takes a SizedLayout and applies it when sizing the IrTensor, but moves the populated tensor into ConstTensorAttributes still carrying the shape derived from the TFLite dims. The two then disagree and ReserveGraphTensors() rejects the model before uploading anything.

TensorDescriptor::UploadData() assigns shape_ from the source tensor and lays the data out against it, so a payload that kept the pre-layout shape would silently discard the requested layout even if the cross-check did not exist.

Writing the laid-out shape back into the tensor is what makes the `layout` parameter mean anything for the data as well as for the graph.

### DIFF
--- a/ml_drift_delegate/tflite/convert/convert_aux.cc
+++ b/ml_drift_delegate/tflite/convert/convert_aux.cc
@@ -62,8 +62,11 @@ template <typename TensorType>
   const TfLiteTensor* tfl_tensor = context.tensors + tensor_id;
   PopulateTensor<TensorType>(tfl_tensor, tensor_id, &t,
                              PopulateTensorFlags::kNoExtraBytes);
-  const ::ml_drift::BHWC shape =
-      GetShape(t.shape, layout, tfl_tensor->dims->size);
+  // Apply the layout to the tensor itself, not just to the IrTensor: the
+  // constant is uploaded from `attr` and its shape is checked against the
+  // IrTensor's, so the two have to agree.
+  t.shape = GetShape(t.shape, layout, tfl_tensor->dims->size);
+  const ::ml_drift::BHWC& shape = t.shape;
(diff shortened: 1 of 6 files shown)


huningxin:
@snnn @terryheo , PTAL, this fix would unblock Chromium side change: https://chromium-review.googlesource.com/c/chromium/src/+/8366452/comment/6270c839_e90cf7da/, WebNN gather constant 1d indices tests failed because of this issue.

/cc @fujunwei @reillyeon

terryheo:
Has started internal review.

terryheo (approved):

huningxin:
@terryheo , as GraphFloat32 path is still used, I appended a commit https://github.com/google-ai-edge/LiteRT/pull/9852/commits/084c864c893cd48cbe300092f228baa5037b9616 to apply the same fixes there as well. Please take another look. Thanks!

huningxin:
@terryheo @snnn , there is an import/copybara issue, I guess it is because I appended a new commit to fix GraphFloat32 path after your approval. Could you please review the new commit? Thanks!

PR #9906, opened 2026-09-11
[MLDrift] Bound the input rank in the CUMSUM support check and make ExtractAxisFromIndex() reject out-of-range axes

wangw-1991:
`IsCumsumSupported()` never validated the rank of `inputs[0]`. It now requires rank 1..5, matching what the IR's BHWDC layout can represent. `ExtractAxisFromIndex()` indexed its lookup table without a bounds check. The table holds `min(rank, 5)` entries, so any axis outside `[-rank, rank)` read past its end. It now returns `Axis::UNKNOWN` instead. Add related tests for the changes.

see chromium issue: https://issues.chromium.org/issues/500050511

### DIFF
--- a/ml_drift_delegate/tflite/convert/convert_aux.cc
+++ b/ml_drift_delegate/tflite/convert/convert_aux.cc
@@ -454,10 +454,13 @@ void PopulateBlockwiseQuantizedFullyConnected(
                      ::ml_drift::Axis::WIDTH, ::ml_drift::Axis::DEPTH,
                      ::ml_drift::Axis::CHANNELS};
   }
-  // The table above tops out at the 5 BHWDC axes, so a tensor of rank 6 or
-  // more can produce an `index` past its end. Return UNKNOWN instead of
-  // indexing out of bounds so that the failure surfaces downstream during
-  // kernel selection rather than as a memory error.
+  // `index_to_axis` has one entry per axis of this tensor, capped at the 5
+  // BHWDC axes, so an out-of-range axis indexes past its end. Callers are
+  // expected to reject those in their Is*Supported() check, but that is
+  // enforced per-op and these checks might be missing in some places,
+  // so this is a form of defense-in-depth. Return UNKNOWN instead of
+  // reading out of bounds so that the failure surfaces during kernel
+  // selection rather than as a memory error.
   if (index < 0 || index >= static_cast<int>(index_to_axis.size())) {
     return ::ml_drift::Axis::UNKNOWN;
   }
--- a/ml_drift_delegate/tflite/convert/convert_aux.cc
+++ b/ml_drift_delegate/tflite/convert/convert_aux.cc
@@ -454,6 +454,13 @@ void PopulateBlockwiseQuantizedFullyConnected(
                      ::ml_drift::Axis::WIDTH, ::ml_drift::Axis::DEPTH,
                      ::ml_drift::Axis::CHANNELS};
   }
+  // The table above tops out at the 5 BHWDC axes, so a tensor of rank 6 or
+  // more can produce an `index` past its end. Return UNKNOWN instead of
+  // indexing out of bounds so that the failure surfaces downstream during
+  // kernel selection rather than as a memory error.
+  if (index < 0 || index >= static_cast<int>(index_to_axis.size())) {
+    return ::ml_drift::Axis::UNKNOWN;
+  }
   return index_to_axis[index];
 }
 
--- a/ml_drift_delegate/tflite/support/support_cumsum.cc
+++ b/ml_drift_delegate/tflite/support/support_cumsum.cc
@@ -25,6 +25,8 @@
 
 namespace litert::ml_drift::ir {
 
+constexpr int kMaxDims = 5;
+
 bool IsCumsumSupported(const TfLiteContext* absl_nonnull context,
                        const TfLiteNode* absl_nonnull node,
                        const TfLiteRegistration* absl_nonnull registration,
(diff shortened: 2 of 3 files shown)


wangw-1991:
@snnn, could you please take a look or help assign some reviewers? Thanks!

ml_drift_delegate/tflite/convert/convert_aux.cc
```diff
@@ -454,6 +454,13 @@ ::ml_drift::Axis ExtractAxisFromIndex(const TfLiteTensor& tflite_tensor,
                      ::ml_drift::Axis::WIDTH, ::ml_drift::Axis::DEPTH,
                      ::ml_drift::Axis::CHANNELS};
   }
+  // The table above tops out at the 5 BHWDC axes, so a tensor of rank 6 or
```
snnn: We don't support 6D tensors in MLDrift.

snnn (changes requested):

ml_drift_delegate/tflite/convert/convert_aux.cc
wangw-1991: Thanks, this is ambiguous; I've updated the comments. Please take a look again.

snnn:
Is it a security fix? If yes, recommend checking with https://github.com/google-ai-edge/LiteRT/blob/main/g3doc/instructions/security.md

Is the issue reachable from WebNN? if not, I will prefer to close this PR.

wangw-1991:
> Is it a security fix? If yes, recommend checking with https://github.com/google-ai-edge/LiteRT/blob/main/g3doc/instructions/security.md
> 
> Is the issue reachable from WebNN? if not, I will prefer to close this PR.

This is a security fix and is reachable from WebNN. Other ops already validate input rank with CheckTensorDims(). CUMSUM missed this check and we need to perform the same checks on it.

snnn:
> Other ops already validate input rank with CheckTensorDims(). CUMSUM missed this check 

Should WebNN do the rank check to ensure all tensors are <= 6 dimensions ?

wangw-1991:
> > Other ops already validate input rank with CheckTensorDims(). CUMSUM missed this check
> 
> Should WebNN do the rank check to ensure all tensors are <= 6 dimensions ?

Thanks, I don't think the check belongs in WebNN. The two layers mean different things by "unsupported":
IsCumsumSupported() returning false just excludes the node from the delegate's partition — LiteRT leaves it to the built-in CPU kernel. WebNN's rank validation, on the other hand, is an API-level hard reject: the graph build fails outright. There is no per-op fallback at that layer. So hoisting an ML Drift GPU limit into WebNN would turn "this CUMSUM runs on CPU" into "the whole graph fails to build."

wangw-1991:
@snnn, Friendly ping.

snnn:
MLDrift does not support >=6D tensors. XNNPack does not support >=7D tensors. Given that, why don't we reject such graphs in the WebNN layer?

wangw-1991:
> MLDrift does not support >=6D tensors. XNNPack does not support >=7D tensors. Given that, why don't we reject such graphs in the WebNN layer?

Thanks. The WebNN layer uses a single shared opSupportLimits for both the XNNPack and MLDrift paths, so a WebNN-level rank gate only has two options, and neither is good:

- Tighten to 5D (the intersection) — this also rejects 6D CumSum that XNNPack handles fine, i.e. a functional regression for the XNNPack path.
- Keep 6D — MLDrift is unprotected and the out-of-bounds path stays reachable, so it doesn't actually address the bug.

The delegate-level check gives exactly the granularity we need instead. IsCumsumSupported() is the IsNodeSupportedFn fed into GraphPartitionHelper, so returning false simply keeps that node out of the MLDrift partition and leaves it on the default TFLite/XNNPack path. Net effect: 5D CumSum runs on GPU, 6D CumSum falls back per-node, and the graph as a whole still runs. A WebNN-level rejection would fail the whole graph build instead.

Besides, the `ToAxis()` change fixes an out-of-bounds read. That shouldn't be reachable for any input regardless of what a higher layer promises, and MLDrift is also consumed directly as a plain TFLite delegate, not only through WebNN.

cc @huningxin @reillyeon

reillyeon:
@snnn, as Wei points out regardless of what WebNN supports, since the higher layers of the LiteRT stack accept tensors with higher rank it is the responsibility of accelerators not to try to delegate operators they don't actually support.

@wangw-1991, I'm not strictly opposed to adjusting the maximum tensor rank that WebNN exposes to developers for contexts where we know we are going to try to use an accelerator/EP which doesn't support a given rank as long as it's greater than the specified minimum required.

wangw-1991:
@snnn, Can you help continue to review this? Thanks.

@reillyeon Thanks. Maybe we can return different maximum tensor ranks for different accelerators/EPs in WebNN. But anyway, it is worth fixing this problem in LiteRT. Because other operators do similar checks, it should be a bug for CumSum to omit this check.

snnn:
> Tighten to 5D (the intersection) — this also rejects 6D CumSum that XNNPack handles fine, i.e. a functional regression for the XNNPack path.

The WebNN backend isn't released yet, so it is not a regression. Do we have a concrete case that there is a valid model which must use 6D tensors?

snnn (changes requested):
If MLDrift does not support 6D tensors at all, we should reject such during delegation, instead of doing op-per-op hardening. I remember the old Graph32 parser has such?

wangw-1991:
> The WebNN backend isn't released yet, so it is not a regression. Do we have a concrete case that there is a valid model which must use 6D tensors?

@snnn You want WebNN to limit the rank to no more than 5, right? I don't know if there is a valid model which must use 6D tensors, but if such a model exists, it will fail to compile on the WebNN LiteRT backend. @reillyeon any comments about this?

> If MLDrift does not support 6D tensors at all, we should reject such during delegation, instead of doing op-per-op hardening. I remember the old Graph32 parser has such?

Yes, old Graph32 parser has such validation, but the ir_model path currently relies on validation by the individual ops themselves. You want similar validation for the `ir_model` path as well, right?

PR #9970, opened 2026-09-15
Fix memory safety vulnerabilities in the GPU delegate model builder

wangw-1991:
Fixes four out-of-bounds accesses that a malformed model can trigger while the GPU delegate builds its graph.

1. `CreateVectorCopyData<float>` copied `src.bytes` into a destination that the caller sized from `NumElements()`, so a tensor whose declared byte count exceeds its shape overflowed the destination buffer; the length check now covers every supported source type.
2. `ObjectReader::GetInputTensor/GetOutputTensor` added `kTfLiteOptionalTensor (-1)` to `context->tensors`, handing callers a pointer before the start of the array that they then dereferenced.
3. `TransposeOperationParser::Parse` used the permutation entries as indices into `index_to_axis` without validating them, so an entry outside `[0, rank)` read past the end of that table.
4. `BatchToSpaceOperationParser::Parse` and `SpaceToBatchOperationParser::Parse` rejected the crop/padding shape with `&&` instead of `||`, letting a 2x1 tensor through before reading data[3].

see chromium issue: https://issues.chromium.org/issues/500050511


tflite/delegates/gpu/common/model_builder.cc
```diff
@@ -2785,6 +2785,15 @@ class TransposeOperationParser : public TFLiteOperationParser {
     TransposeAttributes attr;
     Tensor<Linear, DataType::INT32> perm;
     RETURN_IF_ERROR(reader->ReadTensor(1, &perm));
+    // Entries index `index_to_axis` below, so each one must be a valid axis of
+    // the permutation itself.
+    const int32_t perm_size = static_cast<int32_t>(perm.data.size());
```
snnn, line 2790: Instead of inserting such changes to every backend,  builtin CPU kernels/xnnpack/MLDrift/..., it should be done in a centralized way.

snnn (changes requested):

tflite/delegates/gpu/common/model_builder.cc
wangw-1991: Thanks for the suggestion. Please take a look again.

wangw-1991:
@snnn Can you help review again? Thanks.

PR #10044, opened 2026-09-17
[MLDrift] Read a float16 constant as float32 for a float32 CONCATENATION

huningxin:
The fp16 graph partitioner rewires an mmap'd float16 weight straight into its consuming operation:
FP16GraphPartitionHelper::RemapFp16InputTensors turns fp16 -> DEQUANTIZE -> fp32 -> OP into fp16 -> OP. CONCATENATION takes constants as graph tensors, so the rewrite may leave it with a float16 operand beside float32 ones when requiring float32 precision. The ML Drift CONCAT kernel requires all of its inputs to have the same type and fails to compile such CONCATENATION, e.g. in the WebNN depth-anything-v2-small model [1].

[1]: https://issues.chromium.org/issues/535290311


huningxin:
@snnn @terryheo , PTAL, thanks!

snnn:
Could we store the options, or just `enable_reduced_precision`, by value in `ConcatenationOperationParser`?

The new [`options_` member](https://github.com/google-ai-edge/LiteRT/blob/948225319628858acbf5c54db9c2e4c34d65c4d3/ml_drift_delegate/tflite/model_builder.cc#L1315) is a reference, but the existing [`NewOperationParser` convenience overload](https://github.com/google-ai-edge/LiteRT/blob/948225319628858acbf5c54db9c2e4c34d65c4d3/ml_drift_delegate/tflite/model_builder.cc#L7148) passes temporary options via `/*options=*/{}`. That temporary is destroyed before the factory returns, leaving the returned concat parser with a dangling reference. Calling `Parse()` on that parser then reads `options_.enable_reduced_precision` after the options object's lifetime has ended.

The normal `BuildModel` path keeps its options alive during parsing, so this concerns the convenience factory path. Copying the boolean or the options struct into the parser would fix the lifetime issue.

huningxin:
> The new [`options_` member](https://github.com/google-ai-edge/LiteRT/blob/948225319628858acbf5c54db9c2e4c34d65c4d3/ml_drift_delegate/tflite/model_builder.cc#L1315) is a reference, but the existing [`NewOperationParser` convenience overload](https://github.com/google-ai-edge/LiteRT/blob/948225319628858acbf5c54db9c2e4c34d65c4d3/ml_drift_delegate/tflite/model_builder.cc#L7148) passes temporary options via `/*options=*/{}`.

@snnn , Good catch, thanks. Fix it by holding `ModelBuilderOptions` by value in the latest comment https://github.com/google-ai-edge/LiteRT/pull/10044/commits/a52c98c50ccdee3b9c7a4715508988d3e2ef116c. Please take another look.

